PR #9, opened 2019-04-11
Included Quboo Plugin

mechero:
Added Quboo Plugin properties and also registered in the list.


quboo.properties
```diff
@@ -0,0 +1,14 @@
...
+publicVersions=1.0.3
+
+defaults.mavenGroupId=com.thepracticaldeveloper
+defaults.mavenArtifactId=quboo-sonar-plugin
+
+1.0.3.description=First public version of Quboo Sonar Plugin
+1.0.3.sqVersions=[6.7,LATEST]
+1.0.3.downloadUrl=https://github.com/thepracticaldeveloper/quboo-sonarqube-plugin/releases/download/v1.0.3/quboo-sonar-plugin.jar
+1.0.3.date=2019-04-11
+1.0.3.changelogUrl=https://github.com/thepracticaldeveloper/quboo-sonarqube-plugin/blob/master/README.md
```
ganncamp: This URL should be release-specific, not just point to your README

mechero: I changed it so it points now to the release description.

PR #10, opened 2019-04-24
Fix minimum required versions of 1.5 and 1.6

janos-ss:


janos-ss:
Thanks Ann. That's my understanding too. The purpose of this change is to avoid potential confusion for future readers of the file.

PR #11, opened 2019-04-29
sonar-checkstyle 8.18 & 8.19

muhlba91:


muhlba91:
https://community.sonarsource.com/t/new-release-sonar-checkstyle-8-19/9218

checkstyle.properties
```diff
@@ -1,12 +1,24 @@
...
-archivedVersions=2.2,2.3,2.4,3.5.1,3.6,3.6.1,3.7,4.7,4.8,4.9,4.10,4.11,4.16
-publicVersions=4.17
+archivedVersions=2.2,2.3,2.4,3.5.1,3.6,3.6.1,3.7,4.7,4.8,4.9,4.10,4.11,4.16,4.17,4.18
+publicVersions=4.19
 
 defaults.mavenGroupId=org.sonarsource.checkstyle
 defaults.mavenArtifactId=sonar-checkstyle-plugin
 
+4.19.description=Upgrade to Checkstyle 8.19
+4.19.sqVersion=[6.7,7.7+]
```
ganncamp: This range should end with `LATEST`. As-is, the plugin will be marked incompatible with the imminent 7.8 release.

checkstyle.properties
```diff
@@ -1,12 +1,24 @@
...
+4.19.date=2019-04-29
+4.19.changelogUrl=https://github.com/checkstyle/sonar-checkstyle/releases/tag/4.19
+4.19.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.19/checkstyle-sonar-plugin-4.19.jar
+
+4.18.description=Upgrade to Checkstyle 8.18
```
ganncamp: It's entirely pointless to retroactively add this version. The point of retaining archived versions in these files is to the Marketplace recognizes them as valid versions of the plugin and offers an upgrade to the _current_ version. Since this version goes straight into the archive, no one will have it installed and so there's no need to upgrade _from_ it.

checkstyle.properties
muhlba91: Thank you! :)
I have changed it to `LATEST`.

checkstyle.properties
muhlba91: Thank you for pointing this out. :)
However, `4.18` contained some fixes for SQ 7.6 and, afaik, people installed it manually - hence, I'd like to keep it as an archived version.

PR #12, opened 2019-04-30
Add softvis3d plugin release 1.1.0

stefanrinderle:
SoftVis3D-Plugin is now adapted to the new directory layout concept of SonarQube 7.6. Compatibility versions adapted accordingly.

### DIFF
--- a/softvis3d.properties
+++ b/softvis3d.properties
@@ -2,13 +2,19 @@ category=Visualization/Reporting
 description=Provides 3D models
 homepageUrl=http://softvis3d.com
 archivedVersions=0.5.2,0.6.0,0.7.0,0.8.0,0.8.1,0.9.0,1.0
-publicVersions=1.0.1
+publicVersions=1.1.0,1.0.1
 
 defaults.mavenGroupId=de.rinderle.softvis3d
 defaults.mavenArtifactId=softvis3d
 
+1.1.0.description=Plugin is now adapted to the new directory layout concept.
+1.1.0.sqVersions=[7.6,LATEST]
+1.1.0.date=2019-04-30
+1.1.0.changelogUrl=https://github.com/stefanrinderle/softvis3d/releases/tag/softvis3d-1.1.0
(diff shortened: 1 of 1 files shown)


softvis3d.properties
```diff
@@ -2,13 +2,19 @@ category=Visualization/Reporting
...
 
+1.1.0.description=Plugin is now adapted to the new directory layout concept.
+1.1.0.sqVersions=[7.6,LATEST]
+1.1.0.date=2019-04-30
+1.1.0.changelogUrl=https://github.com/stefanrinderle/softvis3d/releases/tag/softvis3d-1.1.0
+1.1.0.downloadUrl=https://github.com/stefanrinderle/softvis3d/releases/download/softvis3d-1.1.0/sonar-softvis3d-plugin-1.1.0.jar
+
 1.0.1.description=New dashboard features - dark theme and auto reload
-1.0.1.sqVersions=[6.3,LATEST]
+1.0.1.sqVersions=[6.3,7.5]
```
ganncamp: This should really be ended at the current version, 7.7 because users may have installed 1.0.1 in their SQ 7.7 instances. If you retroactively revoke compatibility, I'm not sure what the Marketplace will 'understand' about offering the new version as an upgrade.

Don't worry that that would make 2 versions overlap in compatibility; it's okay because 1.0.1 is archived.

stefanrinderle: Thanks for reviewing this. I've changed the sqVersions for the plugin version 1.0.1 to 7.7 but i didn't understand your comment about the overlapping versions. 1.0.1 will not be archived now as it is needed as public version for all SonarQube instances with version < 7.6. So now I have overlapping SQ versions in as publicVersions stated plugins. As far as I understood the documentation "Compatibility of Public versions cannot overlap". So what to do now?  

I would recommend for you to check how the marketplace will react in such a scenario. As a plugin developer, I cannot always check if a new version of SonarQube has any breaking changes in the plugin before a new SonarQube version is released. This currently means that the version 1.0.1 is stated as compatible with SonarQube 7.7 but in fact it is incompatible. It would be great if the marketplace can handle such a scenario in the future.

PR #15, opened 2019-05-14
sonar-checkstyle 4.20

muhlba91:


muhlba91:
https://community.sonarsource.com/t/new-release-sonar-checkstyle-4-20/9744

checkstyle.properties
```diff
@@ -1,20 +1,26 @@
...
 defaults.mavenArtifactId=sonar-checkstyle-plugin
 
+4.20.description=Upgrade to Checkstyle 8.20
+4.20.sqVersions=[6.7,LATEST]
+4.20.date=2019-05-14
+4.20.changelogUrl=https://github.com/checkstyle/sonar-checkstyle/releases/tag/4.20
+4.20.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.20/checkstyle-sonar-plugin-4.20.jar
+
 4.19.description=Upgrade to Checkstyle 8.19
 4.19.sqVersions=[6.7,LATEST]
```
ganncamp: Only one version can be compatible with `LATEST`

checkstyle.properties
```diff
@@ -1,20 +1,26 @@
...
 4.19.date=2019-04-29
 4.19.changelogUrl=https://github.com/checkstyle/sonar-checkstyle/releases/tag/4.19
 4.19.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.19/checkstyle-sonar-plugin-4.19.jar
 
 4.18.description=Upgrade to Checkstyle 8.18
-4.18.sqVersions=[6.7,7.6+]
+4.18.sqVersions=[6.7,7.7+]
```
ganncamp: No point in this update. If there were, you'd use a `*` instead of a `+`. But archived compatibility is used only for helping users upgrade _from_ these archived versions. By practice and convention this ending range would stay set to the SQ version that was current when this plugin version was archived.

checkstyle.properties
muhlba91: thank you, changed :)

checkstyle.properties
muhlba91: ok, good to know! i have reverted this change

PR #19, opened 2019-05-26
Quboo Plugin v1.0.5: Added support for authenticated servers and bugfix

mechero:
Plugin updated: 
* It now allows entering an API token for servers protected by authentication.
* A bug that prevented the score to be calculated has been fixed.
* Added truststore to properly verify the server certificate.

### DIFF
--- a/quboo.properties
+++ b/quboo.properties
@@ -8,7 +8,7 @@ defaults.mavenGroupId=com.thepracticaldeveloper
 defaults.mavenArtifactId=quboo-sonar-plugin
 
 1.0.3.description=First public version of Quboo Sonar Plugin
-1.0.3.sqVersions=[6.7,LATEST]
+1.0.3.sqVersions=[6.7,7.8]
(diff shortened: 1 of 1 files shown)


quboo.properties
```diff
@@ -1,8 +1,8 @@
 category=Visualization/Reporting
 description=Provides integration with Quboo to use Gamification techniques to fix your legacy code.
 homepageUrl=https://quboo.io
-archivedVersions=
-publicVersions=1.0.3
+archivedVersions=1.0.3
```
ganncamp: I can't comment on the relevant line (#11) so I'll comment here. Only one version can be compatible with `LATEST`, and right now you've got that for both versions. The range end for v1.0.3 needs to be updated. The convention is to set it at the version of SQ that was current at the time of the replacing release, so 7.8 at this point.

mechero: Changed it  ✅

PR #28, opened 2019-07-03
Sonar-PMD 3.2.1

jensgerdes:


pmd.properties
```diff
@@ -1,11 +1,19 @@
...
+3.2.1.description=Improved rule descriptions
+3.2.1.sqVersions=[6.7,LATEST]
+3.2.1.date=2019-04-15
+3.2.1.changelogUrl=https://github.com/jensgerdes/sonar-pmd/releases/tag/3.2.1
+3.2.1.downloadUrl=https://github.com/jensgerdes/sonar-pmd/releases/download/3.2.1/sonar-pmd-plugin-3.2.1.jar
+
+#========================
+
 3.2.0.description=Support SonarQube 7.6+ and PMD 6.10.0
 3.2.0.sqVersions=[6.7,LATEST]
```
ganncamp: Only one version can be compatible with `LATEST`. The convention is to replace `LATEST` with the current SQ version at the time of the change.

PR #31, opened 2019-07-11
Update iCode compatibility to end at 7.8

colin-mueller-sonarsource:


colin-mueller-sonarsource:
See https://community.sonarsource.com/t/plugin-incompatibility-with-jdk-e-g-sonar-i-code-cnes-plugin/11752/3

PR #38, opened 2019-08-20
sonar-mybatis 1.0.2

donhui:
Add sonar.mybatis.skip parameter to skip mybatis sensor

### DIFF
--- a/mybatis.properties
+++ b/mybatis.properties
@@ -2,13 +2,19 @@ category=External Analysers
 description=Rules to check SQL statements in MyBatis Mapper XML files.
 homepageUrl=https://github.com/donhui/sonar-mybatis
 archivedVersions=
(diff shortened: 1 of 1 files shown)


mybatis.properties
```diff
@@ -2,13 +2,19 @@ category=External Analysers
 description=Rules to check SQL statements in MyBatis Mapper XML files.
 homepageUrl=https://github.com/donhui/sonar-mybatis
 archivedVersions=
-publicVersions=1.0.1
+publicVersions=1.0.1,1.0.2
```
ganncamp: public versions can't overlap in compatibility. You need to move 1.0.1 to the archived list.

donhui: I have moved version 1.0.1 to the archived list. Thanks.

PR #41, opened 2019-08-23
Update communitybsl to 1.1.0

nixel2007:


communitybsl.properties
```diff
@@ -1,12 +1,18 @@
...
 defaults.mavenArtifactId=sonar-communitybsl-plugin
 
+1.1.0.description=Cognitive Complexity metric, 9 new rules.
+1.1.0.sqVersions=[7.9,LATEST]
+1.1.0.date=2019-08-23
+1.1.0.changelogUrl=https://github.com/1c-syntax/sonar-bsl-plugin-community/releases/tag/v1.1.0
+1.1.0.downloadUrl=https://github.com/1c-syntax/sonar-bsl-plugin-community/releases/download/v1.1.0/sonar-communitybsl-plugin-1.1.0.jar
+
 1.0.1.description=First release to SonarQube Marketplace.
 1.0.1.sqVersions=[7.9,LATEST]
```
ganncamp: Only one version can be compatible with `LATEST`. The convention is to change the end of the compatibility range to the current version of SQ, but you can certainly go earlier if you like.

PR #49, opened 2019-09-06
[NEW RELEASE] SCM Jazz RTC plugin v1.2

troosan:


scmjazzrtc.properties
```diff
@@ -20,3 +21,9 @@ defaults.mavenArtifactId=sonar-scm-jazzrtc-plugin
 1.1.date=08-06-2015
 1.1.changelogUrl=https://jira.sonarsource.com/secure/ReleaseNote.jspa?projectId=10967&version=11863
 
+1.2.description=Compatibility with SQ 7.7+
+1.2.sqVersions=[7.7,7.7.*],[7.8,7.8.*],[7.9,LATEST]
```
henryju: Why not `[7.7,LATEST]` instead?

troosan: I updated as suggested

PR #50, opened 2019-09-09
Cloudformation Plugin for SonarQube

pethers:
new entry https://community.sonarsource.com/t/new-release-sonar-cloudformation-plugin-1-0-2/14021


cfn.properties
```diff
@@ -0,0 +1,21 @@
...
+defaults.mavenGroupId=com.hack23.sonar
+defaults.mavenArtifactId=sonar-cloudformation-plugin
+
+1.0.2.description=Update dependency com.fasterxml.jackson.core:jackson-databind to v2.10.0.pr2
+1.0.2.date=2019-09-08
+1.0.2.sqVersions=[7.7,LATEST]
+1.0.2.changelogUrl=https://github.com/Hack23/sonar-cloudformation-plugin/releases/tag/sonar-cloudformation-plugin-1.0.2
+1.0.2.downloadUrl=https://oss.sonatype.org/service/local/repositories/releases/content/com/hack23/sonar/sonar-cloudformation-plugin/1.0.2/sonar-cloudformation-plugin-1.0.2.jar
+
+1.0.1.description=New rules added from cfn-nag
```
ganncamp: There's not a lot of point in backfilling this old version, but you can if you want

cfn.properties
```diff
@@ -0,0 +1,21 @@
...
+1.0.1.date=2019-08-17
+1.0.1.sqVersions=[7.7,LATEST]
```
ganncamp: If you are going to include this old version, you must update the compatibility range: only one version can be compatible with `LATEST`. The convention is to set the end of the range to the version of SQ that's current at the time of the release of the plugin replacement.

cfn.properties
```diff
@@ -0,0 +1,21 @@
...
+category=External Analysers
+descripton=Cloudformation rules, supporting cfn-nag
+homepageUrl=https://github.com/Hack23/sonar-cloudformation-plugin
+archivedVersions=1.0.1
+publicVersions=1.0.2
+
+defaults.mavenGroupId=com.hack23.sonar
+defaults.mavenArtifactId=sonar-cloudformation-plugin
+
+1.0.2.description=Update dependency com.fasterxml.jackson.core:jackson-databind to v2.10.0.pr2
```
ganncamp: Especially since this description is purely technical, I'd suggest you use instead something like "Initial Marketplace entry"

cfn.properties
```diff
@@ -0,0 +1,21 @@
...
+1.0.2.date=2019-09-08
+1.0.2.sqVersions=[7.7,LATEST]
+1.0.2.changelogUrl=https://github.com/Hack23/sonar-cloudformation-plugin/releases/tag/sonar-cloudformation-plugin-1.0.2
+1.0.2.downloadUrl=https://oss.sonatype.org/service/local/repositories/releases/content/com/hack23/sonar/sonar-cloudformation-plugin/1.0.2/sonar-cloudformation-plugin-1.0.2.jar
+
+1.0.1.description=New rules added from cfn-nag
+1.0.1.date=2019-08-17
+1.0.1.sqVersions=[7.7,LATEST]
+1.0.1.changelogUrl=https://github.com/Hack23/sonar-cloudformation-plugin/releases/tag/sonar-cloudformation-plugin-1.0.1
+1.0.1.downloadUrl=https://oss.sonatype.org/service/local/repositories/releases/content/com/hack23/sonar/sonar-cloudformation-plugin/1.0.1/sonar-cloudformation-plugin-1.0.1.jar
```
ganncamp: GH seems to want a newline at the end of this line

ganncamp:
In addition to creating your properties file, you must also update https://github.com/SonarSource/sonar-update-center-properties/blob/master/update-center-source.properties by adding your plugin key to the list. In alpha-order please.

cfn.properties
pethers: removed the old version, agree with you.

cfn.properties
pethers: fixed

cfn.properties
pethers: Added a new line

PR #53, opened 2019-09-25
sonar-mybatis 1.0.5

donhui:
- bugfix: handle TypeHandler,javaType,jdbcType,parameterMap,cache,cache-ref when generate reduced.xml
- update mybatis version to 3.5.2

### DIFF
--- a/mybatis.properties
+++ b/mybatis.properties
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Rules to check SQL statements in MyBatis Mapper XML files.
 homepageUrl=https://github.com/donhui/sonar-mybatis
-archivedVersions=1.0.1,1.0.2,1.0.3
-publicVersions=1.0.4
+archivedVersions=1.0.1,1.0.2,1.0.3,1.0.4
+publicVersions=1.0.5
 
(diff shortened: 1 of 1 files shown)


mybatis.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=1.0.1,1.0.2,1.0.3
-publicVersions=1.0.4
+archivedVersions=1.0.1,1.0.2,1.0.3,1.0.4
+publicVersions=1.0.5
 
 defaults.mavenGroupId=org.sonarsource.plugins.mybatis
 defaults.mavenArtifactId=sonar-mybatis-plugin
 
+1.0.5.description=bugfix: handle typeHandler,javaType,jdbcType,parameterMap,cache,cache-ref when generate reduced.xml and update mybatis version to 3.5.2
+1.0.5.sqVersions=[6.7,LATEST]
```
ganncamp: 6.7 is End-of-Life. Please update this to start at 7.9

donhui: > start at 7.9

Thanks for your suggestion, I have updated it.

PR #57, opened 2019-10-12
Update webdriver properties to include new release

kwoding:


webdriver.properties
```diff
@@ -1,12 +1,18 @@
...
 defaults.mavenArtifactId=sonar-webdriver-plugin
 
+1.0.5.description=Upgrade sonar versions to sync with 7.9.1
+1.0.5.sqVersions=[6.7,LATEST]
+1.0.5.date=2019-10-12
+1.0.5.changelogUrl=https://github.com/kwoding/sonar-webdriver-plugin/releases/tag/sonar-webdriver-plugin-1.0.5
+1.0.5.downloadUrl=https://github.com/kwoding/sonar-webdriver-plugin/releases/download/sonar-webdriver-plugin-1.0.5/sonar-webdriver-plugin-1.0.5.jar
+
 1.0.3.description=Add support for SonarQube 6.7
 1.0.3.sqVersions=[6.7,LATEST]
```
ganncamp: Only one version at a time can be compatible with `LATEST`. The convention is to set the old plugin version to the version of SQ that's current at the time, so 7.9.1 in this case, or maybe 7.9.*

kwoding: Ok, updated!

PR #69, opened 2019-11-07
jQAssistant 1.7.1

DirkMahler:
Added entry for jQAssistant plugin


jqassistant.properties
```diff
@@ -0,0 +1,21 @@
+category=External Analyzers
+description=Analyze issues reported by jQAssistant.
+homepageUrl=https://github.com/jqassistant-contrib/sonar-jqassistant-plugin
+archivedVersions=1.7.0
+publicVersions=1.7.1
+ 
+defaults.mavenGroupId=org.jqassistant.contrib.sonarqube
+defaults.mavenArtifactId=sonar-jqassistant-plugin
+
+1.7.1.description=Reduced size of plugin JAR file
```
ganncamp: This is the first version that will be available in the Marketplace. Far better version description would be something along the lines of "Initial entry into the Marketplace"

jqassistant.properties
```diff
@@ -0,0 +1,21 @@
...
+1.7.1.date=2019-11-10
+1.7.1.sqVersions=[7.9,LATEST]
+1.7.1.changelogUrl=https://github.com/jqassistant-contrib/sonar-jqassistant-plugin/blob/master/release-notes.adoc#171
+1.7.1.downloadUrl=https://search.maven.org/remotecontent?filepath=org/jqassistant/contrib/sonarqube/sonar-jqassistant-plugin/1.7.1/sonar-jqassistant-plugin-1.7.1.jar
+
+1.7.0.description=Support for SonarQube 7.9 LTS and Maven multi-module projects
```
ganncamp: There's almost no point in including this. We keep archived version data so the Marketplace can recognize when upgrades are available. Since no one has installed this version through the Mktplce, you've got a vanishingly small upgrade audience here. But this is up to you.

jqassistant.properties
```diff
@@ -0,0 +1,21 @@
...
+1.7.0.date=2019-11-07
+1.7.0.sqVersions=[7.9,LATEST]
```
ganncamp: Only one version can be marked compatible with `LATEST`. The convention is to mark the version being archived as ending compatibility with the version of SQ that's current at the time.

DirkMahler:
Applied changes by reducing to initial entry of release 1.7.1 (i.e. completely removing 1.7.0)

PR #78, opened 2019-12-19
Initial entry of the sonar-devon4j-plugin into the marketplace

lmarniazman:
See our topic on the community forum: 
https://community.sonarsource.com/t/new-release-sonar-devon4j-plugin-3-2-1/17962


devon4j.properties
```diff
@@ -0,0 +1,14 @@
+category=Governance
+description=Plugin for SonarQube to validate devon4j architecture and conventions.
```
ganncamp: The "Plugin for SonarQube to" part is a bit redundant in this context, but okay.

devon4j.properties
```diff
@@ -0,0 +1,14 @@
...
+homepageUrl=https://github.com/devonfw/sonar-devon4j-plugin/#sonar-devon4j-plugin
+archivedVersions=
+publicVersions=3.2.1
+
+defaults.mavenGroupId=com.devonfw.tools
+defaults.mavenArtifactId=sonar-devon4j-plugin
+
+3.2.1.description=First official release under the name sonar-devon4j-plugin and initial entry into the marketplace.
+3.2.1.sqVersions=[6.4,7.*]
```
ganncamp: This range needs to be updated to reflect current SQ versions/support. Assuming compatibility, I suggest: `[7.9,LATEST]`
Note that only one version at a time can be marked compatible with latest, so on your second release you'll need to update this version. But using `LATEST` has the benefit that as new versions of SQ are released, your most recent plugin version is automatically marked compatible.

devon4j.properties
```diff
@@ -0,0 +1,14 @@
...
+3.2.1.date=2019-12-17
+3.2.1.changelogUrl=https://github.com/devonfw/sonar-devon4j-plugin/blob/master/CHANGELOG.adoc#321
+3.2.1.downloadUrl=https://repo.maven.apache.org/maven2/com/devonfw/tools/sonar-devon4j-plugin/3.2.1/sonar-devon4j-plugin-3.2.1.jar
```
ganncamp: GH appears to want a newline at the end here...

lmarniazman:
Thanks for the review! 
I updated the compatible versions and changed the wording of the version description, as it was a little off.

PR #81, opened 2019-12-23
Ansible plugin versions 2.2.0 and 2.3.0

sbaudoin:


ansible.properties
```diff
@@ -1,12 +1,24 @@
...
-publicVersions=2.1.0
+archivedVersions=1.0.0,2.0.0,2.2.0
+publicVersions=2.3.0
  
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=ansible
 
+2.3.0.description=Support for SonarQube 8
+2.3.0.date=2019-12-23
+2.3.0.sqVersions=[6.7,LATEST]
```
ganncamp: Only one version can be compatible with `LATEST`

ansible.properties
```diff
@@ -1,12 +1,24 @@
...
+2.3.0.changelogUrl=https://github.com/sbaudoin/sonar-ansible/releases/tag/v2.3.0
+2.3.0.downloadUrl=https://github.com/sbaudoin/sonar-ansible/releases/download/v2.3.0/sonar-ansible-plugin-2.3.0.jar
+
+2.2.0.description=Ability to use and supply custom ansible-lint configuration files
+2.2.0.date=2019-05-30
+2.2.0.sqVersions=[6.7,7.9]
+2.2.0.changelogUrl=https://github.com/sbaudoin/sonar-ansible/releases/tag/v2.2.0
+2.2.0.downloadUrl=https://github.com/sbaudoin/sonar-ansible/releases/download/v2.2.0/sonar-ansible-plugin-2.2.0.jar
```
ganncamp: There is no point in adding this directly into the archived section. Either make it public and compatible with a non-overlapping subset of SQ versions or remove it.

sbaudoin: Version 2.2 was actually missing in this file (I forgot to create the PR for this version) so I added it back for consistency reasons

ansible.properties
ganncamp: Could you bump the start of this range up to 7.9, please? Sorry I didn't notice that the first time around.

sbaudoin:
Done

ansible.properties
```diff
@@ -1,15 +1,27 @@
...
-publicVersions=2.1.0
+archivedVersions=1.0.0,2.0.0,2.1.0,2.2.0
+publicVersions=2.3.0
  
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=ansible
 
+2.3.0.description=Support for SonarQube 8
+2.3.0.date=2019-12-23
+2.3.0.sqVersions=[7.9,LATEST]
```
ganncamp: This description says "Support for SQ 8". Should compatibility start there?

sbaudoin: 7 and 8 are both supported with this version. I meant here that this is the first version that supports SQ 8 but not exclusively.

PR #83, opened 2020-01-10
Move 6.7 - 7.8 to private versions

colin-mueller-sonarsource:
SonarQube 6.7 through 7.8 are all considered EOL now that six months have passed since the new LTS was released.

Services Team discussed this and we are fully comfortable with this since no plugin updates have been released for 6.7 - 7.8 since the new LTS was released in July.

### DIFF
--- a/update-center-source.properties
+++ b/update-center-source.properties
@@ -35,25 +35,225 @@ ltsVersion=7.9.2
 7.9.date=2019-07-01
 
 #==================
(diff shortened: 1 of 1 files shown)


colin-mueller-sonarsource:
@christophelevis In this file or in others? In this file -- 4.5 - 6.7.4 were listed in the order of oldest version to newest version before I touched anything.

christophelevis:
@colin-mueller-sonarsource It concerns all files. Prior to a certain date, we were used to add new versions at the bottom. We decided to change it, but haven't reordered existing versions.
If we want to clean up, the order would then rather be: latest version first :)

colin-mueller-sonarsource:
@christophelevis reordered :)

PR #84, opened 2020-01-10
Release of Sonargraph Plugin 4.1.1

ingmarkellner:


sonargraphintegration.properties
```diff
@@ -1,21 +1,21 @@
...
 defaults.mavenGroupId=com.hello2morrow
 defaults.mavenArtifactId=sonar-sonargraph-integration
 
-4.0.0.description=Single execution of plugin, compatibility with SonarQube 7.9.1
-4.0.0.sqVersions=[6.7.6,LATEST]
-4.0.0.date=2019-08-29
-4.0.0.changelogUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/tag/Release-4.0.0_with_releasenotes
-4.0.0.downloadUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/download/Release-4.0.0_with_releasenotes/sonar-sonargraph-integration-4.0.0.jar
+4.1.1.description=Compatibility with SonarQube 8.1
+4.1.1.sqVersions=[6.7.6,LATEST]
```
ganncamp: At a minimum I need you to bump the start of the range to 7.9; support for 6.7 is official over. Also, the description says 8.1 compatibility; should this range actually start there?

ingmarkellner: The minimum version is now set to 7.9.
This release fixes a problem that was introduced with more restrictive security settings in SQ8, leading to classloading problems (https://community.sonarsource.com/t/access-denied-java-lang-runtimepermission-getclassloader-for-plugin-in-computing-engine-of-sq-8-0-and-sq-8-1/17916). Thus, no, compatibility does not start there, but includes the latest version.

sonargraphintegration.properties
```diff
@@ -1,21 +1,21 @@
...
 defaults.mavenGroupId=com.hello2morrow
 defaults.mavenArtifactId=sonar-sonargraph-integration
 
-4.0.0.description=Single execution of plugin, compatibility with SonarQube 7.9.1
-4.0.0.sqVersions=[6.7.6,LATEST]
-4.0.0.date=2019-08-29
-4.0.0.changelogUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/tag/Release-4.0.0_with_releasenotes
-4.0.0.downloadUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/download/Release-4.0.0_with_releasenotes/sonar-sonargraph-integration-4.0.0.jar
+4.1.1.description=Compatibility with SonarQube 8.1
+4.1.1.sqVersions=[7.9,LATEST]
```
ganncamp: The comment says compatibility with 8.1. Shouldn't the compatibility range start there?

ingmarkellner: The description means that the plugin runs now also with SonarQube 8.1, which did not work before. The plugin is still compatible with the lower versions of SonarQube.

PR #86, opened 2020-01-23
Archive old plugin versions, fix compatibility ranges

christophelevis:


christophelevis:
Thanks @ganncamp. I'll do it in a second time then.

PR #91, opened 2020-02-05
Release sonargraph plugin 4.1.2

ingmarkellner:


sonargraphintegration.properties
```diff
@@ -4,20 +4,26 @@ category=External Analyzers
...
+4.1.2.description=Compatibility with Sonargraph 9.13 and newer
+4.1.2.sqVersions=[7.9,LATEST]
+4.1.2.date=2020-02-05
+4.1.2.changelogUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/download/Release-4.1.2/release-notes.txt
+4.1.2.downloadUrl=https://github.com/sonargraph/sonar-sonargraph-integration/releases/download/Release-4.1.2/sonar-sonargraph-integration-4.1.2.jar
+
+#====================
 4.1.1.description=Compatibility with SonarQube 8.1
-4.1.1.sqVersions=[7.9,LATEST]
+4.1.1.sqVersions=[6.7,8.1]
```
ganncamp: 6.7 is no longer recognized in the Marketplace. Please keep the bottom of this range at 7.9.

ingmarkellner: I thought this does not matter for archived versions?

ganncamp: a) we've archived versions of SQ <7.9 so IIRC, this will cause errors in the generation job
b) it started out at 7.9 & compatibility on archived versions is about upgrade prompts in the Marketplace. Since this wasn't originally marked compatible with 6.7, no one on 6.7 should be running the version & _need_ to upgrade from it.

ingmarkellner: Never mind, I thought, since the plugin version 4.1.1 was programmed against the older API, it would be ok to mention it in the archived section. Wasn't aware of the consequences.

PR #93, opened 2020-02-06
SCM-TFVC 2.3.0

lg2de:
*switch to Visual Studio 2019 using Microsoft.TeamFoundationServer.ExtendedClient
*fix user identification when connecting using Live ID
*check configuration for collection URI
*introduce build information

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target/
+/.vs
--- a/scmtfvc.properties
+++ b/scmtfvc.properties
@@ -2,8 +2,8 @@ category=Integration
 description=Provides SCM TFVC integration
 devVersion=
 privateVersions=
-publicVersions=2.3,2.2
-archivedVersions=1.0,2.0,2.1,2.1.1,2.1.2
+publicVersions=2.3
+archivedVersions=1.0,2.0,2.1,2.1.1,2.1.2,2.2
 
 defaults.mavenGroupId=org.sonarsource.sonar-plugins.tfvc
 defaults.mavenArtifactId=sonar-scm-tfvc-plugin
--- a/scmtfvc.properties
+++ b/scmtfvc.properties
@@ -9,13 +9,13 @@ defaults.mavenGroupId=org.sonarsource.sonar-plugins.tfvc
 defaults.mavenArtifactId=sonar-scm-tfvc-plugin
 
 2.3.description=embed TFS API
-2.3.sqVersions=[7.3,LATEST]
+2.3.sqVersions=[7.9,LATEST]
 2.3.downloadUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/download/2.3.0/sonar-scm-tfvc-plugin-2.3.0-73.jar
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
 target/
+/.vs
```
ganncamp: What is this change? I probably need you to remove it.

scmtfvc.properties
```diff
@@ -2,12 +2,18 @@ category=Integration
...
 defaults.mavenArtifactId=sonar-scm-tfvc-plugin
 
+2.3.description=embed TFS API
+2.3.sqVersions=[7.3,LATEST]
+2.3.downloadUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/download/2.3.0/sonar-scm-tfvc-plugin-2.3.0-73.jar
+2.3.date=2020-02-06
+2.3.changelogUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/tag/2.3.0
+
 2.2.description=Fix compatibility for SQ 7.3 and above
 2.2.sqVersions=[7.3,LATEST]
```
ganncamp: Only one version can be compatible with `LATEST` at a time. Best practice is to set the old version's range end to the current version of SQ.

scmtfvc.properties
```diff
@@ -2,12 +2,18 @@ category=Integration
...
 privateVersions=
-publicVersions=2.2
+publicVersions=2.3,2.2
 archivedVersions=1.0,2.0,2.1,2.1.1,2.1.2
 
 defaults.mavenGroupId=org.sonarsource.sonar-plugins.tfvc
 defaults.mavenArtifactId=sonar-scm-tfvc-plugin
 
+2.3.description=embed TFS API
+2.3.sqVersions=[7.3,LATEST]
```
ganncamp: SQ versions <7.9 have been archived. Please start this range at 7.9.

lg2de: Mmh, I think it would work also with older versions.
Why do you want to stop it?
Anyway, I've changed it.

.gitignore
lg2de: This is related to a folder created by Visual Studio.
Adding this exclusion the project can also be edited with Visual Studio.

scmtfvc.properties
```diff
@@ -2,14 +2,20 @@ category=Integration
 description=Provides SCM TFVC integration
 devVersion=
 privateVersions=
-publicVersions=2.2
+publicVersions=2.3,2.2
```
ganncamp: Sorry, I didn't catch the first time that you weren't archiving the old version. In fact, I need you to archive the old version because the compatibility of public versions can't overlap. Normally, you'd set the end range of 2.2 to before 2.3 compatibility starts, but that puts it ending at 7.8 and _that_ SQ version has been archived. So... the upshot is that you need to move 2.2 to `archivedVersions`.

lg2de: done

PR #97, opened 2020-02-19
New request to publish icode plugin 2.0.1

begarco:
## Modification
- Fix pluginKey and artifactId in icode.properties

## Plugin changelog

- i-Code analyzer is included directly in the plugin, there is no more need to install and launch i-Code
- analysis is now recursive
- support for java 11
- enhancement of the CI for the plugin

### DIFF
--- a/icode.properties
+++ b/icode.properties
@@ -1,11 +1,17 @@
 category=External Analyzers
-description=Import i-Code CNES results for Fortran(77 & 90) and Shell
+description=Analyze your Fortran(77 & 90) and Shell code with i-Code CNES
 homepageUrl=https://github.com/lequal/sonar-icode-cnes-plugin
 archivedVersions=1.0.2,1.1.0
-publicVersions=
+publicVersions=2.0.0
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=sonaricode
+
+2.0.0.description=i-Code CNES is now embedded in the plugin, no more need to install it
+2.0.0.date=2020-02-14
+2.0.0.sqVersions=[6.7.7,8.1]
+2.0.0.changelogUrl=https://github.com/lequal/sonar-icode-cnes-plugin/releases/tag/2.0.0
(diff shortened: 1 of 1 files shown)


icode.properties
```diff
@@ -1,11 +1,19 @@
 category=External Analyzers
-description=Import i-Code CNES results for Fortran(77 & 90) and Shell
+description=Analyze your Fortran(77 & 90) and Shell code with i-Code CNES
 homepageUrl=https://github.com/lequal/sonar-icode-cnes-plugin
 archivedVersions=1.0.2,1.1.0
-publicVersions=
+publicVersions=2.0.1
+
+sonar.pluginKey=icode
```
ganncamp: This belongs in your project pom, not this file. As extraneous data, it should be removed at your next release, but I don't think it will interfere with this release.

icode.properties
```diff
@@ -1,11 +1,19 @@
...
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
-defaults.mavenArtifactId=sonaricode
+defaults.mavenArtifactId=sonar-icode-cnes-plugin
+
+2.0.1.description=i-Code CNES is now embedded in the plugin, no more need to install it
+2.0.1.date=2020-02-14
+2.0.1.sqVersions=[7.9,8.1]
```
ganncamp: The end range of the current version is typically set to `LATEST`. By setting the end of the range to 8.1, you've already marked yourself with last week's 8.2 release.

icode.properties
begarco: Ok we will remove it in the next PR, I open an issue on our project to keep it in mind.

icode.properties
begarco: We had a problem with our cnesreport plugin: we set the sqVersions with [6.7,LATEST] but the plugin is no more compatible with SonarQube 8 and users were able to download it on version 8 before we were aware of the project. 

If you prefer we use LATEST, we can.

ganncamp: Not if it's not compatible. :-)
I'm just pointing out that with the compatibility you've set here, the plugin isn't offered to users of SonarQube 8.2

icode.properties
ganncamp: Just to make sure: you don't remove it in _your_ project. You _need_ it in your project (presumably). You remove it from _this file._

PR #98, opened 2020-02-20
Build Breaker 2.3.1

sandro-h:
Corresponding community thread: https://community.sonarsource.com/t/requesting-inclusion-into-marketplace-for-sonar-build-breaker/20427


buildbreaker.properties
```diff
@@ -0,0 +1,13 @@
+category=Integration
+description=Break the analyzer if the project does not pass its Quality Gate
+homepageUrl=https://github.com/adnovum/sonar-build-breaker
+archivedVersions=
+publicVersions=2.3.1
+defaults.mavenGroupId=org.sonarqubecommunity.buildbreaker
+defaults.mavenArtifactId=sonar-build-breaker-plugin
+
+2.3.1.description=General cleanup & inclusion in SQ Marketplace
+2.3.1.sqVersions=[7.3, LATEST]
```
ganncamp: Versions below 7.9 are archived. This range needs to start at 7.9

buildbreaker.properties
```diff
@@ -0,0 +1,13 @@
+category=Integration
+description=Break the analyzer if the project does not pass its Quality Gate
```
ganncamp: Technically, this "fails the analysis" rather than _breaking_ the _analyzer_. An update here would be nice.

buildbreaker.properties
sandro-h: Thanks, I updated the range.

buildbreaker.properties
sandro-h: Fair point, the wording is a bit strange. I updated it to your suggestion.

PR #99, opened 2020-02-23
codehawk rules v1.0

Jcseeco:


codehawk.properties
```diff
@@ -0,0 +1,14 @@
...
+description=Analyze Java code smell
+homepageUrl=https://github.com/SDPMLab/CodeHawk
+archivedVersions=1.0
+publicVersions=1.0
+
+defaults.mavenGroupId= org.codehawk.plugin
+defaults.mavenArtifactId= codehawk
+
+1.0.description= Initial version compatible with SonarQube 7.7
+1.0.sqVersions=[7.7,LATEST]
```
ganncamp: This should start at 7.9, the oldest version listed in the Marketplace

Jcseeco: Done, but our plugin still need fixing.

Jcseeco:
I've updated the file, thank you!

codehawk.properties
```diff
@@ -1,26 +1,14 @@
 category=External Analysers
 description=Analyze Java code smell
 homepageUrl=https://github.com/SDPMLab/CodeHawk
-archivedVersions=1.0
-publicVersions=1.0
+archivedVersions=1.2
```
ganncamp: It can't be both public and archived. You'll (probably) _move_ 1.2 to this list once you release 1.3

Jcseeco: Removed value of archivedVersions

PR #101, opened 2020-02-26
add OverOps plugin to Marketplace

daverted:


overops.properties
```diff
@@ -0,0 +1,14 @@
...
+description=Identify and drill down into runtime errors in Java applications
+homepageUrl=https://github.com/takipi-field/sonar-plugin-overops
+archivedVersions=
+publicVersions=1.0.3
+
+defaults.mavenGroupId=com.overops.plugins.sonar
+defaults.mavenArtifactId=overops-plugin
+
+1.0.3.description=Initial Release
+1.0.3.sqVersions=[7.9,LATEST]
```
ganncamp: Your docs say compatibility starts at 7.9.1...?

overops.properties
```diff
@@ -0,0 +1,14 @@
...
+1.0.3.date=2020-02-24
+1.0.3.changelogUrl=https://github.com/takipi-field/sonar-plugin-overops/releases/tag/v1.0.3
+1.0.3.downloadUrl=https://github.com/takipi-field/sonar-plugin-overops/releases/download/v1.0.3/overops-plugin-1.0.3.jar
```
ganncamp: Not a big deal, but GitHub seems to want a linebreak at the end.

overops.properties
daverted: Yes, thanks! I've confirmed the correct version is 7.9.1 and updated it here.

overops.properties
daverted: 👍

PR #106, opened 2020-03-11
Add Sonar CNES Report 3.2.2

Sancretor:
# Changelog
* Support SonarQube 8 up to 8.2
* Support Java 11
* Compatibility with branch feature

### DIFF
--- a/cnesreport.properties
+++ b/cnesreport.properties
@@ -14,6 +14,6 @@ defaults.mavenArtifactId=cnesreport
 3.2.2.downloadUrl=https://github.com/lequal/sonar-cnes-report/releases/download/3.2.2/sonar-cnes-report-3.2.2.jar
 3.1.0.description=Support SonarQube 7.9 LTS and release to marketplace
 3.1.0.date=2019-08-21
-3.1.0.sqVersions=[6.7,7.9]
+3.1.0.sqVersions=[6.7,8.2]
 3.1.0.changelogUrl=https://github.com/lequal/sonar-cnes-report/releases/tag/3.1.0
 3.1.0.downloadUrl=https://github.com/lequal/sonar-cnes-report/releases/download/3.1.0/sonar-cnes-report-3.1.0.jar
--- a/cnesreport.properties
+++ b/cnesreport.properties
@@ -1,14 +1,19 @@
 category=Visualization/Reporting
 description=CNES plugin that allows users to download a bundle of project reports in multiple formats.
 homepageUrl=https://github.com/lequal/sonar-cnes-report
-archivedVersions=
-publicVersions=3.1.0
+archivedVersions=3.1.0
+publicVersions=3.2.2
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
+3.2.2.description=Support SonarQube 8 up to 8.2 & add ability to export from a specific branch
+3.2.2.date=2020-03-10
+3.2.2.sqVersions=[7.9,8.2]
+3.2.2.changelogUrl=https://github.com/lequal/sonar-cnes-report/releases/tag/3.2.2
+3.2.2.downloadUrl=https://github.com/lequal/sonar-cnes-report/releases/download/3.2.2/sonar-cnes-report-3.2.2.jar
 3.1.0.description=Support SonarQube 7.9 LTS and release to marketplace
 3.1.0.date=2019-08-21
-3.1.0.sqVersions=[6.7,LATEST]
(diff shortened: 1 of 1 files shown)


cnesreport.properties
```diff
@@ -1,14 +1,19 @@
...
 
+3.2.2.description=Support SonarQube 8 up to 8.2 & add ability to export from a specific branch
+3.2.2.date=2020-03-10
+3.2.2.sqVersions=[7.9,8.2]
+3.2.2.changelogUrl=https://github.com/lequal/sonar-cnes-report/releases/tag/3.2.2
+3.2.2.downloadUrl=https://github.com/lequal/sonar-cnes-report/releases/download/3.2.2/sonar-cnes-report-3.2.2.jar
 3.1.0.description=Support SonarQube 7.9 LTS and release to marketplace
 3.1.0.date=2019-08-21
-3.1.0.sqVersions=[6.7,LATEST]
+3.1.0.sqVersions=[6.7,7.9]
```
ganncamp: The convention is to set this to the version that's current at the time. If you set this to 7.9 then people using CNES 3.1.0 on SQ 8.2 won't get an upgrade suggestion in the Marketplace.

cnesreport.properties
```diff
@@ -1,14 +1,19 @@
...
-publicVersions=3.1.0
+archivedVersions=3.1.0
+publicVersions=3.2.2
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
+3.2.2.description=Support SonarQube 8 up to 8.2 & add ability to export from a specific branch
+3.2.2.date=2020-03-10
+3.2.2.sqVersions=[7.9,8.2]
```
ganncamp: This range should end with `LATEST` unless you specifically know that this version _won't_ be compatible with 8.3. That way when SQ 8.3 is released this version will automatically be marked as compatible with it.

cnesreport.properties
```diff
@@ -1,14 +1,19 @@
...
 homepageUrl=https://github.com/lequal/sonar-cnes-report
-archivedVersions=
-publicVersions=3.1.0
+archivedVersions=3.1.0
+publicVersions=3.2.2
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
+3.2.2.description=Support SonarQube 8 up to 8.2 & add ability to export from a specific branch
```
ganncamp: See my comment about versions & forward compatibility. Suggest

> Support SonarQube 8.0+

cnesreport.properties
Sancretor: Actually, we submit a new version mostly because the previous one does not work on SonarQube 8.x... but people can install it because of the LATEST keyword.
If I get this right, I should revert this change so that people on SQ 8 can upgrade to the new version ? Correct ?

cnesreport.properties
Sancretor: Like I said, the previous version is not compatible with SQ 8, and because of the LATEST keyword, we received a lot of issues and we were not able to deal with them fast enough.  
For SQ 8.3, we can't predict if a change would break our plugin.  

So now, we would like to have control on the compatible versions... that's why I ended the range with the last SQ release.  
Is that still a problem ?

cnesreport.properties
ganncamp: Ah! Incompatible as in doesn't work or stops the server from spinning up? All new installs will be offered the latest version regardless of what's listed here. So unless we're talking about preventing server startup, this should _probably_ be at 8.2 to make sure anyone who has it installed on 8.0-8.2 is offered the upgrade.
You _do not_ want to revert this to `LATEST` because only one version can be listed as compatible with it at a time.

cnesreport.properties
ganncamp: Totally up to you. Let's leave it at 8.2. AND we still don't want to revert the previous version to `LATEST`.

cnesreport.properties
Sancretor: See my previous comment on `sqVersions`.

ganncamp: Given our other conversations, I'm resolving this one.

cnesreport.properties
Sancretor: No server issues, the plugin is just not able to reach a specific API.  
Thanks your explanations, now I understand how this works.
I'll change 7.9 for 8.2. 
Thanks !

cnesreport.properties
Sancretor: Ok, let's keep 8.2 then.  
And I still won't revert the previous version to LATEST, I promise !

PR #112, opened 2020-04-02
Release: AEMRules v1.1

jplucinski:
**AEMRules v1.1**

Release notes:
- Java Plugin 6.2 support

### DIFF
--- a/aemrules.properties
+++ b/aemrules.properties
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Adds rules for AEM Java development
 homepageUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube
-archivedVersions=0.8,0.9,0.10,0.11
-publicVersions=1.0
+archivedVersions=0.8,0.9,0.10,0.11,1.0
+publicVersions=1.1
 
 defaults.mavenGroupId=com.cognifide.aemrules
 defaults.mavenArtifactId=sonar-aemrules-plugin
 
+1.1.description=Java Plugin 6.2 support
+1.1.sqVersions=[7.9,LATEST]
+1.1.date=2020-04-02
+1.1.changelogUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/tag/v1.1
+1.1.downloadUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/download/v1.0/sonar-aemrules-plugin-1.0.jar
+
 1.0.description=HTL Support
-1.0.sqVersions=[7.9,LATEST]
+1.0.sqVersions=[7.9,8.2]
 1.0.date=2020-02-07
 1.0.changelogUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/tag/v1.0
 1.0.downloadUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/download/v1.0/sonar-aemrules-plugin-1.0.jar


aemrules.properties
```diff
@@ -1,14 +1,20 @@
...
 
+1.1.description=Java Plugin 6.2 support
+1.1.sqVersions=[7.9,LATEST]
+1.1.date=2020-04-02
+1.1.changelogUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/tree/v1.1
+1.1.downloadUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/download/v1.0/sonar-aemrules-plugin-1.0.jar
+
 1.0.description=HTL Support
-1.0.sqVersions=[7.9,LATEST]
+1.0.sqVersions=[7.9,8.1]
```
ganncamp, line 17: You really want to set this to 8.2 so people on that version can be prompted to update.

aemrules.properties
```diff
@@ -1,14 +1,20 @@
...
+archivedVersions=0.8,0.9,0.10,0.11,1.0
+publicVersions=1.1
 
 defaults.mavenGroupId=com.cognifide.aemrules
 defaults.mavenArtifactId=sonar-aemrules-plugin
 
+1.1.description=Java Plugin 6.2 support
+1.1.sqVersions=[7.9,LATEST]
+1.1.date=2020-04-02
+1.1.changelogUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/tree/v1.1
```
ganncamp, line 13: This doesn't go to a change log. Did you mean this instead?
https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/tag/v1.1

jplucinski: Thanks! Fixed.

aemrules.properties
jplucinski: Changed to 8.2.

aemrules.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=0.8,0.9,0.10,0.11
-publicVersions=1.0
+archivedVersions=0.8,0.9,0.10,0.11,1.0
+publicVersions=1.1
 
 defaults.mavenGroupId=com.cognifide.aemrules
 defaults.mavenArtifactId=sonar-aemrules-plugin
 
+1.1.description=Java Plugin 6.2 support
+1.1.sqVersions=[7.9,LATEST]
```
ganncamp, line 11: Now that I'm actually _reading_ your release notes, is this the correct range minimum? Your notes list 7.9.2 as the SQ min. (versus 7.9[.0]) and v6.0 of the Java analyzer. SonarQube 8.1 shipped with a 5.x version & 8.2 shipped with v6.1

jplucinski: Hi,
I fixed release notes, there should be 7.9.0. Issue about versions of Java analyser is that SQ forces to update plugin on server restart. So, if I install AEM Rules 1.0 on SQ 7.9.x, after server restart, I’ll have AEM Rules 1.0 and Java analyser 6.2, this configuration fails while scanning code.  That’s why I want to AEM Rules 1.1 to support Java analyser 6.x. Of course, if someone want to install AEM Rules manually via jar file on SQ with 5.x Java analyser should use AEM 1.0.
I don’t exactly know if this is right solution for this issue.

Thanks for your analysis.

ganncamp: We can't explicitly manage dependency versions here. That's handled based on the dependencies you declare. So I think we're good enough here.

PR #113, opened 2020-04-06
Sonargraph Integration Plugin Version 5.0.0

ingmarkellner:
Makes the Sonargraph Integration Plugin compatible with Sonargraph 9.14.0 and later.


ingmarkellner:
Need to do some modification...

PR #116, opened 2020-04-08
Announcement for sonar-apigee-plugin 2.1.0 release

ntiss:


apigee.properties
```diff
@@ -2,11 +2,17 @@ category=External Analyzers
...
 defaults.mavenArtifactId=sonar-apigee-plugin
 
+2.1.0.description=New features
+2.1.0.sqVersions=[6.7,LATEST]
+2.1.0.date=2020-04-08
+2.1.0.changelogUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin/releases/tag/v2.1.0
+2.1.0.downloadUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin/releases/download/v2.1.0/sonar-apigee-plugin-2.1.0.jar
+
 2.0.0.description=Support SonarXML 2.0.1+
 2.0.0.sqVersions=[6.7,LATEST]
```
ganncamp, line 17: You can't have two versions compatible with `LATEST`. You need to set the end range for the one you're archiving to the current version of SQ: 8.2 in this case.

apigee.properties
```diff
@@ -2,11 +2,17 @@ category=External Analyzers
...
 homepageUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin
 archivedVersions=1.1.0,1.2.0,1.3.0
-publicVersions=2.0.0
+publicVersions=2.1.0
 
 defaults.mavenGroupId=com.arkea.satd.sonar.apigee
 defaults.mavenArtifactId=sonar-apigee-plugin
 
+2.1.0.description=New features
+2.1.0.sqVersions=[6.7,LATEST]
```
ganncamp, line 11: 7.9 is the oldest version still supported in the Marketplace. Please start the range for this plugin version there.

apigee.properties
```diff
@@ -2,11 +2,17 @@ category=External Analyzers
 description=Adds XML rules test Apigee apiproxies.
 homepageUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin
 archivedVersions=1.1.0,1.2.0,1.3.0
-publicVersions=2.0.0
+publicVersions=2.1.0
 
 defaults.mavenGroupId=com.arkea.satd.sonar.apigee
 defaults.mavenArtifactId=sonar-apigee-plugin
 
+2.1.0.description=New features
```
ganncamp, line 10: Really? Surely you want to give your users more than this?
Suggest: SharedFlow support, new rules.

apigee.properties
ntiss: Done.

apigee.properties
ntiss: You're right. Done.

apigee.properties
ntiss: Problem here, my company still uses SQ 7.6.
Is it possible to make an exception and start the range at 7.6?

ganncamp: This is not about policy, but about how the metadata generator that consumes these files works. However, regardless of the range here, you can install the plugin manually in your company's local instance without problems.

ntiss: I agree. (And we'll install manually). Fix is done.

PR #118, opened 2020-04-09
OIDC Auth 2.0.0

tjuerge:
This release introduces several breaking changes:
* starting with SonarQube 8.0 the [`UserInfo` field `login` is for internal use only](https://github.com/SonarSource/sonarqube/blob/eacba88db0a44e34156e9e31d315ef31d61e726a/sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/UserIdentity.java#L166) (causing failed logins from this plugin)
  **A new SonarQube user's login is always auto-generated by SonarQube. So the plugin's login strategy is only used for the user's *provider* login.**
* instead storing the provider metadata in the plugin's configuration the metadata is now retrieved from the provider
  **This change introduces an additional network trip to the provider during every login.**
* the list of OAuth scopes (even the always required "openid") used by the plugin is fully configurable now


authoidc.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=1.0.1,1.0.2,1.0.3,1.0.4
-publicVersions=1.1.0
+archivedVersions=1.0.1,1.0.2,1.0.3,1.0.4,1.1.0
+publicVersions=2.0.0
 
 defaults.mavenGroupId=org.vaulttec.sonarqube.auth.oidc
 defaults.mavenArtifactId=sonar-auth-oidc-plugin
 
+2.0.0.description=Breaking changes: user's login is always auto-generated by SonarQube, provider metadata is retrieved from the provider and the list of OAuth scopes is fully configurable
+2.0.0.sqVersions=[7.4,LATEST]
```
ganncamp, line 11: This range needs to start at 7.9, the oldest version still supported by the Marketplace.

tjuerge: fixed

PR #121, opened 2020-04-11
Include detekt plugin

arturbosch:


detekt.properties
```diff
@@ -0,0 +1,15 @@
+category=External Analysers
+description=Provide detekt rules for Kotlin projects
+homepageUrl=https://github.com/detekt/sonar-kotlin
+archivedVersions=
+publicVersions=2.0.0
+
+defaults.mavenGroupId=io.github.detekt
+defaults.mavenArtifactId=sonar-detekt
+
+2.0.0.description=Upgrade to detekt 1.8.0
```
ganncamp, line 10: How about "Initial Marketplace entry" instead?

PR #123, opened 2020-04-26
update checkstyle 4.31

muhlba91:
https://community.sonarsource.com/t/new-release-sonar-checkstyle-4-31/23688

### DIFF
--- a/checkstyle.properties
+++ b/checkstyle.properties
@@ -14,7 +14,7 @@ defaults.mavenArtifactId=sonar-checkstyle-plugin
 4.31.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.31/checkstyle-sonar-plugin-4.31.jar
 
 4.30.description=Upgrade to Checkstyle 4.30
-4.30.sqVersions=[7.9,8.1*]
+4.30.sqVersions=[7.9,8.2*]
 4.30.date=2020-03-10
 4.30.changelogUrl=https://github.com/checkstyle/sonar-checkstyle/releases/tag/4.30
 4.30.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.30/checkstyle-sonar-plugin-4.30.jar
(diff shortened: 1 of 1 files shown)


checkstyle.properties
```diff
@@ -1,14 +1,20 @@
...
 
+4.31.description=Upgrade to Checkstyle 4.31
+4.31.sqVersions=[7.9,LATEST]
+4.31.date=2020-04-26
+4.31.changelogUrl=https://github.com/checkstyle/sonar-checkstyle/releases/tag/4.31
+4.31.downloadUrl=https://github.com/checkstyle/sonar-checkstyle/releases/download/4.31/checkstyle-sonar-plugin-4.31.jar
+
 4.30.description=Upgrade to Checkstyle 4.30
-4.30.sqVersions=[7.9,LATEST]
+4.30.sqVersions=[7.9,8.1*]
```
ganncamp, line 17: @muhlba91 the current version is 8.2. To allow current 8.2 users to upgrade through the Marketplace you should end this range there.

In the end it's up to you, though. I'll publish like this if you really want me to.

muhlba91: thank you!

just for my understanding: this means upgrades are only possible for a range that ends with the latest version and only then a marketplace upgrade is guaranteed?

out of curiosity. what would happen if a user upgrades to 8.3 and leaves the old checkstyle version there. is he still able to upgrade through marketplace?

ganncamp: >  this means upgrades are only possible for a range that ends with the latest version and only then a marketplace upgrade is guaranteed?

I don't quite follow this. The ranges on archived versions are about when/how the user will be prompted to upgrade. 

Let's say I'm running SQ 8.0 with Checkstyle 4.29. The Marketplace will prompt me to upgrade to the latest compatible version: 4.31. 

Now let's say I'm running 8.3 with a manually installed 4.29 (because I copied my plugins over when I did my server upgrade [slaps hand]). Since that box in the compatibility matrix isn't checked off, the Marketplace won't understand that 4.31 is a suitable replacement, and I'll never be prompted to upgrade. (This is my current understanding.)

If that doesn't make sense or if you have more questions, let's continue the conversation in the community - since I'm about to merge/close this PR. :)

muhlba91: thank you, now I got it what you meant. :)

PR #129, opened 2020-06-02
sonar-detekt 2.1.0

arturbosch:


detekt.properties
```diff
@@ -1,14 +1,20 @@
...
 
+2.1.0.description=Upgrade to detekt 1.9.1
+2.1.0.sqVersions=[7.9,LATEST]
+2.1.0.date=2020-06-02
+2.1.0.changelogUrl=https://github.com/detekt/sonar-kotlin/releases/tag/2.1.0
+2.1.0.downloadUrl=https://github.com/detekt/sonar-kotlin/releases/download/2.1.0/sonar-detekt-2.1.0.jar
+
 2.0.0.description=Initial Marketplace entry
-2.0.0.sqVersions=[7.9,LATEST]
+2.0.0.sqVersions=[7.9,8.2]
```
ganncamp, line 17: Unless you know of a specific incompatibility with 8.3, you want to set the end range to that. Why? Because users might be on 8.3 and this version of your plugin. If the compatibility range doesn't include 8.3 then those users won't be prompted to upgrade in the Marketplace.

detekt.properties
```diff
@@ -1,14 +1,20 @@
...
+archivedVersions=2.0.0
+publicVersions=2.1.0
 
 defaults.mavenGroupId=io.github.detekt
 defaults.mavenArtifactId=sonar-detekt
 
+2.1.0.description=Upgrade to detekt 1.9.1
+2.1.0.sqVersions=[7.9,LATEST]
+2.1.0.date=2020-06-02
+2.1.0.changelogUrl=https://github.com/detekt/sonar-kotlin/releases/tag/2.1.0
```
ganncamp, line 13: As-is, this changelog isn't terribly helpful. Perhaps you could edit your project tag to add the link to the release notes you posted in your community request?

detekt.properties
arturbosch: Is it recommend to still write `LATEST` here or should I update the numbers to the latest released sonarqube version after I've verified it still works with it?

detekt.properties
arturbosch: Thx, will do!

detekt.properties
```diff
@@ -1,12 +1,18 @@
...
+2.1.0.downloadUrl=https://github.com/detekt/sonar-kotlin/releases/download/2.1.0/sonar-detekt-2.1.0.jar
+
 2.0.0.description=Initial Marketplace entry
 2.0.0.sqVersions=[7.9,LATEST]
```
ganncamp, line 17: Only one version can be compatible with `LATEST`. You should set this to 8.3.

arturbosch: done.

PR #141, opened 2020-08-07
Release: AEMRules v1.2

toniedzwiedz:


aemrules.properties
```diff
@@ -1,14 +1,20 @@
...
 
+1.2.description=Stability improvements, removal of Guava dependencies.
+1.2.sqVersions=[7.9,LATEST]
+1.2.date=2020-08-07
+1.2.changelogUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/tag/v1.2
+1.2.downloadUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube/releases/download/v1.2/sonar-aemrules-plugin-1.2.jar
+
 1.1.description=Java Plugin 6.2 support
-1.1.sqVersions=[7.9,LATEST]
+1.1.sqVersions=[7.9,8.2]
```
ganncamp, line 17: This should end with the current version: 8.4.1 (or 8.4.*). Otherwise your current users on 8.3-8.4.1 won't be prompted to upgrade

toniedzwiedz: Changed to 8.4.*

A prompt to install the new version sounds good, although, depending on other plugins installed, 1.1 could crash at startup. This doesn't happen when running SQ from the Jar file, it does when running it using a docker image. This seems to affect 7.9 and 8.* alike. We decided to remove the dependency that kept causing conflicts (a specific Guava version), which is what the 1.2 release is mostly about. 

As I understand, the current range will work for those users of 1.1 who got it working, which is what we want to happen. What about those who haven't installed the plugin yet? Will they be prompted to install 1.2 (effectively skipping 1.1)

PR #142, opened 2020-08-10
Update xanitizer.properties

normanwenzel:
Added version 2.1.0

### DIFF
--- a/xanitizer.properties
+++ b/xanitizer.properties
@@ -14,7 +14,7 @@ defaults.mavenArtifactId=sonar-xanitizer-plugin
 2.1.0.downloadUrl=https://github.com/RIGS-IT/sonar-xanitizer/releases/download/v2.1.0/sonar-xanitizer-plugin-2.1.0.jar
 
 2.0.0.description=Support for SonarQube 7.3, Integration into Security Reports, Import All option
-2.0.0.sqVersions=[7.3,LATEST]
+2.0.0.sqVersions=[7.3,7.5]
 2.0.0.date=2018-11-13
 2.0.0.changelogUrl=https://github.com/RIGS-IT/sonar-xanitizer/releases/tag/2.0.0
 2.0.0.downloadUrl=https://github.com/RIGS-IT/sonar-xanitizer/releases/download/2.0.0/sonar-xanitizer-plugin-2.0.0.jar
--- a/xanitizer.properties
+++ b/xanitizer.properties
@@ -8,13 +8,13 @@ defaults.mavenGroupId=com.rigsit
 defaults.mavenArtifactId=sonar-xanitizer-plugin
 
 2.1.0.description=Support for SonarQube 7.9, Added support for JavaScript and TypeScript, Added several new Java problem types
-2.1.0.sqVersions=[7.6,LATEST]
+2.1.0.sqVersions=[7.9,LATEST]
 2.1.0.date=2020-08-10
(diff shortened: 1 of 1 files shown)


xanitizer.properties
```diff
@@ -1,14 +1,20 @@
...
 
+2.1.0.description=Support for SonarQube 7.9, Added support for JavaScript and TypeScript, Added several new Java problem types
+2.1.0.sqVersions=[7.6,LATEST]
+2.1.0.date=2020-08-10
+2.1.0.changelogUrl=https://github.com/RIGS-IT/sonar-xanitizer/releases/tag/v2.1.0
+2.1.0.downloadUrl=https://github.com/RIGS-IT/sonar-xanitizer/releases/download/v2.1.0/sonar-xanitizer-plugin-2.1.0.jar
+
 2.0.0.description=Support for SonarQube 7.3, Integration into Security Reports, Import All option
-2.0.0.sqVersions=[7.3,LATEST]
+2.0.0.sqVersions=[7.3,7.5]
```
ganncamp, line 17: This needs to end at the current version, 8.4.1, or users on more recent versions won't be prompted to upgrade

xanitizer.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=1.3.2,1.3.3,1.4.0,1.5.0
-publicVersions=2.0.0
+archivedVersions=1.3.2,1.3.3,1.4.0,1.5.0,2.0.0
+publicVersions=2.1.0
 
 defaults.mavenGroupId=com.rigsit
 defaults.mavenArtifactId=sonar-xanitizer-plugin
 
+2.1.0.description=Support for SonarQube 7.9, Added support for JavaScript and TypeScript, Added several new Java problem types
+2.1.0.sqVersions=[7.6,LATEST]
```
ganncamp, line 11: This needs to start at 7.9, the current LTS.

xanitizer.properties
```diff
@@ -1,14 +1,20 @@
...
 homepageUrl=https://github.com/RIGS-IT/sonar-xanitizer
-archivedVersions=1.3.2,1.3.3,1.4.0,1.5.0
-publicVersions=2.0.0
+archivedVersions=1.3.2,1.3.3,1.4.0,1.5.0,2.0.0
+publicVersions=2.1.0
 
 defaults.mavenGroupId=com.rigsit
 defaults.mavenArtifactId=sonar-xanitizer-plugin
 
+2.1.0.description=Support for SonarQube 7.9, Added support for JavaScript and TypeScript, Added several new Java problem types
```
ganncamp, line 10: Release notes mention Xanitizer 5.0. I don't remember the details (& too lazy to look) so honest question here: did you upgrade an underlying analyzer here? I.e. should this description mention upgrading to X5.0? Or does the user run that independently & import the reports?

normanwenzel: > Release notes mention Xanitizer 5.0. I don't remember the details (& too lazy to look) so honest question here: did you upgrade an underlying analyzer here? I.e. should this description mention upgrading to X5.0? Or does the user run that independently & import the reports?

Dear Ann,

Xanitizer is an external analyzer (which the user runs independently from SQ) and the plugin parses its report and creates issues from it. So I think there is no need to reference this in the description, especially since it is not mandatory to update, because the plugin also supports older versions.

Best regards,
Norman

PR #143, opened 2020-08-14
new release v1.4

Jcseeco:
updated properties for new release v1.4

### DIFF
--- a/codehawk.properties
+++ b/codehawk.properties
@@ -1,20 +1,20 @@
 category=External Analysers
 description=Analyze Java code smell
 homepageUrl=https://github.com/SDPMLab/CodeHawk
-archivedVersions=
-publicVersions=1.3
+archivedVersions=1.3
+publicVersions=1.4
 
 defaults.mavenGroupId= org.codehawk.plugin
 defaults.mavenArtifactId= codehawk
-
-1.3.description= Initial release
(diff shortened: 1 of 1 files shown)


codehawk.properties
```diff
@@ -12,3 +12,9 @@ defaults.mavenArtifactId= codehawk
 1.3.date=2020-03-28
 1.3.downloadUrl=https://github.com/SDPMLab/CodeHawk/releases/download/1.3/codehawk-1.3.jar
 1.3.changelogUrl=https://github.com/SDPMLab/CodeHawk/releases/tag/1.3
+
```
ganncamp, line 15: I can't comment on the relevant lines because they're not in the PR.

First, you can't have 2 versions compatible with `LATEST` at the same time, so on line 11, you need to set the end of the compatibility range to either `8.4.1` or `8.4.*`

Second, you've added 1.4 to this file, but you haven't made it public on line 5. What you need to do is move v1.3 from `publicVersions` on line 5 to `archivedVersions` on line 4, and then make 1.4 public by listing it on line 5.

To be clear, multiple versions _can_ be public at the same time if they don't have overlapping compatibility, but that's not what we're looking at here.

Jcseeco: Hi, I've updated the file.

PR #150, opened 2020-09-17
Release sonar-findbugs 4.0.1

KengoTODA:
We've released [sonar-findbugs version 4.0.1](https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.1), please add it to the marketplace.
Here is [my post at forum for this change](https://community.sonarsource.com/t/new-release-sonar-findbugs-4-0-1/31461).

Thanks in advance!

### DIFF
--- a/findbugs.properties
+++ b/findbugs.properties
@@ -1,13 +1,19 @@
 category=External Analysers
 description=Provide Findbugs rules for analysis of Java projects
-archivedVersions=2.4,3.0,3.1,3.2,3.3,3.4.3,3.4.4,3.5,3.6,3.7,3.8.0,3.9.0,3.9.1,3.9.2,3.9.3,3.10.0,3.11.0,3.9.4,3.11.1
-publicVersions=4.0.0
+archivedVersions=2.4,3.0,3.1,3.2,3.3,3.4.3,3.4.4,3.5,3.6,3.7,3.8.0,3.9.0,3.9.1,3.9.2,3.9.3,3.10.0,3.11.0,3.9.4,3.11.1,4.0.0
+publicVersions=4.0.1
 
 defaults.mavenGroupId=org.sonarsource.sonar-findbugs-plugin
 defaults.mavenArtifactId=sonar-findbugs-plugin
 
+4.0.1.description=Use SpotBugs 4.1.2, sb-contrib 7.4.7, and findsecbugs 1.10.1
+4.0.1.sqVersions=[7.9,LATEST]
+4.0.1.date=2020-09-18
+4.0.1.changelogUrl=https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.1
+4.0.1.downloadUrl=https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.0.1/sonar-findbugs-plugin-4.0.1.jar
+
 4.0.0.description=Use SpotBugs 4.0.0, sb-contrib 7.4.7, and findsecbugs 1.10.1
-4.0.0.sqVersions=[7.9,LATEST]
+4.0.0.sqVersions=[7.9,8.3]
 4.0.0.date=2020-02-28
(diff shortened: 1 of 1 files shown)


findbugs.properties
```diff
@@ -1,13 +1,19 @@
...
 
+4.0.1.description=Use SpotBugs 4.1.2, sb-contrib 7.4.7, and findsecbugs 1.10.1
+4.0.1.sqVersions=[7.9,LATEST]
+4.0.1.date=2020-09-18
+4.0.1.changelogUrl=https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.1
+4.0.1.downloadUrl=https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.0.1/sonar-findbugs-plugin-4.0.1.jar
+
 4.0.0.description=Use SpotBugs 4.0.0, sb-contrib 7.4.7, and findsecbugs 1.10.1
-4.0.0.sqVersions=[7.9,LATEST]
+4.0.0.sqVersions=[7.9,8.3]
```
ganncamp, line 16: This range needs to end with the current version. Otherwise your users on 8.4.* won't be prompted to upgrade in the Marketplace.

findbugs.properties
```diff
@@ -1,13 +1,19 @@
...
+archivedVersions=2.4,3.0,3.1,3.2,3.3,3.4.3,3.4.4,3.5,3.6,3.7,3.8.0,3.9.0,3.9.1,3.9.2,3.9.3,3.10.0,3.11.0,3.9.4,3.11.1,4.0.0
+publicVersions=4.0.1
 
 defaults.mavenGroupId=org.sonarsource.sonar-findbugs-plugin
 defaults.mavenArtifactId=sonar-findbugs-plugin
 
+4.0.1.description=Use SpotBugs 4.1.2, sb-contrib 7.4.7, and findsecbugs 1.10.1
+4.0.1.sqVersions=[7.9,LATEST]
+4.0.1.date=2020-09-18
+4.0.1.changelogUrl=https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.1
```
ganncamp, line 12: The changes in your release tag (version in pom & readme) don't match your description here.

ganncamp (changes requested):

findbugs.properties
KengoTODA: This version has compatibility problem with latest version so I wanna exclude it. Is it impossible?

findbugs.properties
KengoTODA: Sorry but I cannot get your point. Which version in the pom file?

https://github.com/spotbugs/sonar-findbugs/blob/78fe4818b3fccbd6a9fe6ca3982322abd0ec8b5b/pom.xml

ganncamp: Okay, I misread the interface. Pom updated _since_ the tag. Which means that your "change log" gives 0 information about the changes. No narrative description, no list of (or link to) tickets, no commits. Can you please beef up the page this link points to with ... _something_?

findbugs.properties
ganncamp: Not impossible. And okay to do if you don't think anyone will be using the combination. But if you do have someone on SQ8.4 & FB 4.0.0, then you _want_ the Marketplace to prompt an upgrade. Because right now Marketplace is offering 4.0.0 as an install to folks on 8.4.
Also, don't worry that setting the end range to 8.4.* means that folks will be offered 4.0.0 after 4.0.1 is released. Only the latest compatible version of a plugin is ever offered via the Marketplace.

findbugs.properties
KengoTODA: OK I've updated that, please confirm https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.1

ganncamp: Better! Thanks!

ganncamp (approved):

PR #154, opened 2020-10-01
Release sonar-detekt 2.3.0

arturbosch:


detekt.properties
```diff
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Provide detekt rules for Kotlin projects
 homepageUrl=https://github.com/detekt/sonar-kotlin
-archivedVersions=2.0.0
-publicVersions=2.1.0,2.2.0
+archivedVersions=2.0.0,2.2.0
+publicVersions=2.1.0,2.3.0
```
arturbosch, line 5: Note: sonar-detekt 2.2.0 updated the sonar-api to 7.9.4 so 2.1.0 is still a public version for users of sonar < 7.9.4.

detekt.properties
```diff
@@ -1,14 +1,20 @@
...
 
+2.3.0.description=Upgrade to detekt 1.14.1
+2.3.0.sqVersions=[7.9.4,LATEST]
+2.3.0.date=2020-10-01
+2.3.0.changelogUrl=https://github.com/detekt/sonar-kotlin/releases/tag/2.3.0
+2.3.0.downloadUrl=https://github.com/detekt/sonar-kotlin/releases/download/2.3.0/sonar-detekt-2.3.0.jar
+
 2.2.0.description=Upgrade to detekt 1.11.0
-2.2.0.sqVersions=[7.9.4,LATEST]
+2.2.0.sqVersions=[7.9.4,8.4]
```
ganncamp, line 17: This needs to end with the current version (8.4.2) so that users of that version (and below) will be prompted by the Marketplace to upgrade. As-is, users on 8.4.1, 8.4.2 won't be.

detekt.properties
ganncamp: thanks for the reminder! :)

ganncamp (changes requested):

detekt.properties
arturbosch: Done.

ganncamp (approved):

PR #159, opened 2020-10-23
AEM Rules Plugin v1.3 Release

toniedzwiedz:
Bring back Java 8 runtime and build support by popular demand

### DIFF
--- a/aemrules.properties
+++ b/aemrules.properties
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Adds rules for AEM Java development
 homepageUrl=https://github.com/Cognifide/AEM-Rules-for-SonarQube
-archivedVersions=0.8,0.9,0.10,0.11,1.0,1.1
-publicVersions=1.2
+archivedVersions=0.8,0.9,0.10,0.11,1.0,1.1,1.2
+publicVersions=1.3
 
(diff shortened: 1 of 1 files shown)


aemrules.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=0.8,0.9,0.10,0.11,1.0,1.1
-publicVersions=1.2
+archivedVersions=0.8,0.9,0.10,0.11,1.0,1.1,1.2
+publicVersions=1.3
 
 defaults.mavenGroupId=com.cognifide.aemrules
 defaults.mavenArtifactId=sonar-aemrules-plugin
 
+1.3.description=Brought back Java 8 compatibility to allow the rules to be used outside SonarQube Server.
+1.3.sqVersions=[7.9,LATEST]
```
ganncamp, line 11: Your announcement in the community lists compatibility starting at 7.9.3, not 7.9. Which is correct?

ganncamp (changes requested):

aemrules.properties
toniedzwiedz: This hasn't changed compared to the latest version. It should be 7.9. I copied 7.9.3 from the previous announcement since that was the lowest 7.9.* version I had explicitly tested.

ganncamp (approved):

PR #162, opened 2020-11-05
REL-1728

christophe-havard-sonarsource:


christophe-havard-sonarsource:
As you can see, I changed the order of downloads for .NET scanners : now I'd like to priviledge .NET 5, then .NET Core global tool, then .NET Core 2.0 and then .NET Framework 4.6.
Also, I named the new ".NET" as ".NET 5" to avoid confusion. It follows a conversation with @mickael-caro-sonarsource , please feel free to feedback.

PR #187, opened 2021-01-02
Update codehawk.properties

SEPMLAB:
Revised sqVersion from 'latest' to 8.4.*

### DIFF
--- a/codehawk.properties
+++ b/codehawk.properties
@@ -8,7 +8,7 @@ defaults.mavenGroupId= org.codehawk.plugin
 defaults.mavenArtifactId= codehawk
 
 1.4.description= Added new rules
-1.4.sqVersions=[7.9,LATEST]
+1.4.sqVersions=[7.9,8.4.*]
 1.4.date=2020-08-14
 1.4.downloadUrl=https://github.com/SDPMLab/CodeHawk/releases/download/1.4/codehawk-1.4.jar
 1.4.changelogUrl=https://github.com/SDPMLab/CodeHawk/releases/tag/1.4
--- a/codehawk.properties
+++ b/codehawk.properties
@@ -1,20 +1,20 @@
 category=External Analysers
 description=Analyze Java code smell
 homepageUrl=https://github.com/SDPMLab/CodeHawk
-archivedVersions=1.3
-publicVersions=1.4
+archivedVersions=1.4
+publicVersions=1.5
 
 defaults.mavenGroupId= org.codehawk.plugin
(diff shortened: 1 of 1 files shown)


codehawk.properties
```diff
@@ -8,7 +8,7 @@ defaults.mavenGroupId= org.codehawk.plugin
 defaults.mavenArtifactId= codehawk
 
 1.4.description= Added new rules
-1.4.sqVersions=[7.9,LATEST]
+1.4.sqVersions=[7.9,8.4.*]
```
ganncamp, line 11: When you replace `LATEST` on the version you're archiving, you should always replace it with the current version. Otherwise your users on the latest SQ version won't be prompted in-app to upgrade to _your_ latest version.

ganncamp (changes requested):

codehawk.properties
```diff
@@ -1,20 +1,20 @@
...
+1.5.date=2021-01-16
+1.5.downloadUrl=https://github.com/SEPMLAB/CodeHawk/releases/download/V1.5/codehawk-1.5.jar
+1.5.changelogUrl=https://github.com/SEPMLAB/CodeHawk/releases/tag/V1.5
+
 1.4.description= Added new rules
 1.4.sqVersions=[7.9,8.4.*]
 1.4.date=2020-08-14
 1.4.downloadUrl=https://github.com/SDPMLab/CodeHawk/releases/download/1.4/codehawk-1.4.jar
 1.4.changelogUrl=https://github.com/SDPMLab/CodeHawk/releases/tag/1.4
-
```
ganncamp, line 15: Best practice would leave this in place. People get lazy and you shouldn't assume that users have upgraded off of this version already. If you remove this version, people still on it won't be prompted to upgrade.

codehawk.properties
```diff
@@ -1,20 +1,20 @@
...
-archivedVersions=1.3
-publicVersions=1.4
+archivedVersions=1.4
+publicVersions=1.5
 
 defaults.mavenGroupId= org.codehawk.plugin
 defaults.mavenArtifactId= codehawk
 
+1.5.description= Added new rules
+1.5.sqVersions=[7.9,8.4.*]
```
ganncamp, line 11: Do you really mean to keep users of more current versions from using your plugin? The current version is 8.6 (with 8.6.1 on the way). Best practice would end this range with `LATEST` to stay compatible with newest releases.

ganncamp (changes requested):

ganncamp (approved):

PR #186, opened 2021-01-02
Release sonar-findbugs-plugin 4.0.3

KengoTODA:
We've released [sonar-findbugs-plugin 4.0.3](https://github.com/spotbugs/sonar-findbugs/releases/tag/4.0.3), please add it to the marketplace.
  I'll post to the forum and add its URL here later.

  Thanks in advance!

PR #193, opened 2021-01-18
Format the properties file for automation

KengoTODA:
To apply automation to the deployment process, format the properties file by this PR, then #186 can focus on the necessary content update for deploy.

### DIFF
--- a/findbugs.properties
+++ b/findbugs.properties
@@ -1,145 +1,120 @@
-category=External Analysers
-description=Provide Findbugs rules for analysis of Java projects
-archivedVersions=2.4,3.0,3.1,3.2,3.3,3.4.3,3.4.4,3.5,3.6,3.7,3.8.0,3.9.0,3.9.1,3.9.2,3.9.3,3.10.0,3.11.0,3.9.4,3.11.1,4.0.0,4.0.1
-publicVersions=4.0.2
-
-defaults.mavenGroupId=org.sonarsource.sonar-findbugs-plugin
-defaults.mavenArtifactId=sonar-findbugs-plugin
-
-4.0.2.description=Use SpotBugs 4.2.0, sb-contrib 7.4.7, and findsecbugs 1.11.0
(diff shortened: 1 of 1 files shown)


ganncamp (changes requested):
Every single line in this file has changes, compared to the current head version.

KengoTODA:
Yes, I want to format the properties file to ease my work.
To verify that the change won't affect SQ market place, please use the following Java code:

```java
import java.io.BufferedInputStream;
import java.io.BufferedReader;
import java.io.IOException;
import java.net.URI;
import java.net.URLConnection;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.StandardOpenOption;
import java.util.Properties;

public class Main {
    private static final String BEFORE = "https://github.com/SonarSource/sonar-update-center-properties/raw/2667f752414c2cf6e7a80b6d9efb2df87e5f1f87/findbugs.properties";
    private static final String AFTER = "https://github.com/SonarSource/sonar-update-center-properties/raw/2a2419f1d65b0ec473f3f94d69525e01b174ae5b/findbugs.properties";

    private static Path download(String uri) throws IOException {
        URLConnection con = URI.create(uri).toURL().openConnection();
        try (BufferedInputStream input = new BufferedInputStream(con.getInputStream())) {
            byte[] b = input.readAllBytes();
            Path path = Files.createTempFile("", ".properties");
            Files.write(path, b, StandardOpenOption.WRITE);
            return path;
        }
    }

    public static void main(String[] args) throws IOException {
        Properties before = new Properties();
        Properties after = new Properties();
        try (BufferedReader beforeReader = Files.newBufferedReader(download(BEFORE));
             BufferedReader afterReader = Files.newBufferedReader(download(AFTER))) {
            before.load(beforeReader);
            after.load(afterReader);

            System.out.printf("Are these two properties same?: %s%n", before.equals(after));
        }
    }
}
```

ganncamp (approved):

PR #197, opened 2021-02-12
Add Sonar CNES Report plugin 3.3.0

diegorodriguez31:
# Changelog
Fix some issues
Change some display
The plugin is now working for non-admin users
Fix security hotspots display
SonarCloud will not support Java 8 anymore
It is now possible to use plugin mode on sonar running on a port different than 9000
An infinite number of projects are now displayed on the cnes-report page

And more... (https://github.com/cnescatlab/sonar-cnes-report/releases/tag/3.3.0)


Sancretor:
@ganncamp I think this PR can be closed, as the next version (3.3.1) is already available on the marketplace.

PR #205, opened 2021-04-08
Update CNES Report plugin properties

Sancretor:
# Changelog
* Enable use of SonarQube context URL
* Fix date command line argument
* Fix version command line argument
 
And more... (https://github.com/cnescatlab/sonar-cnes-report/releases/tag/3.3.1)

Please note that the previous 3.3.0 request did not receive any answer, so this PR contains version 3.3.0 data too.

### DIFF
--- a/cnesreport.properties
+++ b/cnesreport.properties
@@ -1,7 +1,7 @@
 category=Visualization/Reporting
 description=CNES plugin that allows users to download a bundle of project reports in multiple formats.
 homepageUrl=https://github.com/cnescatlab/sonar-cnes-report
-archivedVersions=3.1.0,3.2.2,3.3.0
+archivedVersions=3.1.0,3.2.2
 publicVersions=3.3.1
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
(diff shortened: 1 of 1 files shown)


cnesreport.properties
```diff
@@ -1,12 +1,22 @@
...
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
+3.3.1.description=Fix issues on date & version command line arguments and add compatibility with SonarQube context
+3.3.1.date=2021-04-07
+3.3.1.sqVersions=[7.9,8.2]
+3.3.1.changelogUrl=https://github.com/cnescatlab/sonar-cnes-report/releases/tag/3.3.1
+3.3.1.downloadUrl=https://github.com/cnescatlab/sonar-cnes-report/releases/download/3.3.1/sonar-cnes-report-3.3.1.jar
+3.3.0.description=Resolve bugs. Add display enhancement, new management of security hotspots and allow non-admins users to use the plugin
```
ganncamp, line 15: No need to add a version straight to the archives. I know I apparently missed a request for this version & I guess you're trying to present a complete change log to the user (kudos!) but we're facing an internal push to clean these files up. Can we assume that the skipped number will prompt users to dig around in GH releases? Or maybe add a note at the end of the release message for 3.3.1?

ganncamp (changes requested):

cnesreport.properties
Sancretor: You guessed right !
Alright, I'll remove the 3.3.0 from the PR.
And I think we already have a note inside the GH release page regarding 3.3.0.

ganncamp (approved):

PR #211, opened 2021-05-10
Change minimum SonarQube version of opendge plugin

gquerret:
... as I incorrectly understood the meaning of the sqVersions property.

### DIFF
--- a/openedge.properties
+++ b/openedge.properties
@@ -2,18 +2,19 @@ category=Languages
 description=Code Analyzer for OpenEdge
 homepageUrl=https://github.com/Riverside-Software/sonar-openedge
 archivedVersions=
-publicVersions=2.13.1,2.14.4
+publicVersions=2.14.4
+archivedVersions=2.13.1
 defaults.mavenGroupId=eu.rssw.sonar.openedge
 defaults.mavenArtifactId=sonar-openedge-plugin
 
 2.14.4.description=New expression engine in ABL Parser
-2.14.4.sqVersions=[7.9,9.1]
+2.14.4.sqVersions=[8.9,LATEST]
 2.14.4.date=2021-10-04
 2.14.4.changelogUrl=https://github.com/Riverside-Software/sonar-openedge/releases/tag/V2.14.4
 2.14.4.downloadUrl=https://github.com/Riverside-Software/sonar-openedge/releases/download/V2.14.4/sonar-openedge-plugin-2.14.4.jar
 
 2.13.1.description=Initial Marketplace entry. Support for OpenEdge 12.3 and 12.4
-2.13.1.sqVersions=[7.6,8.9]
+2.13.1.sqVersions=[7.6,8.2]
 2.13.1.date=2021-05-06
 2.13.1.changelogUrl=https://github.com/Riverside-Software/sonar-openedge/releases/tag/V2.13.1
 2.13.1.downloadUrl=https://github.com/Riverside-Software/sonar-openedge/releases/download/V2.13.1/sonar-openedge-plugin-2.13.1.jar
--- a/openedge.properties
+++ b/openedge.properties
@@ -2,12 +2,18 @@ category=Languages
 description=Code Analyzer for OpenEdge
 homepageUrl=https://github.com/Riverside-Software/sonar-openedge
 archivedVersions=
-publicVersions=2.13.1
+publicVersions=2.13.1,2.14.4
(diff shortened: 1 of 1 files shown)


openedge.properties
```diff
@@ -2,12 +2,18 @@ category=Languages
 description=Code Analyzer for OpenEdge
 homepageUrl=https://github.com/Riverside-Software/sonar-openedge
 archivedVersions=
-publicVersions=2.13.1
+publicVersions=2.13.1,2.14.4
 defaults.mavenGroupId=eu.rssw.sonar.openedge
 defaults.mavenArtifactId=sonar-openedge-plugin
 
+2.14.4.description=New expression engine in ABL Parser
+2.14.4.sqVersions=[7.9,9.1]
```
ganncamp, line 10: 7.9 is fully past EOL. This range needs to start with 8.9.

Additionally, are you sure you don't want to end the range with `LATEST`? 9.2.1 is already out, and this range excludes those users.

openedge.properties
```diff
@@ -2,12 +2,18 @@ category=Languages
...
+2.14.4.date=2021-10-04
+2.14.4.changelogUrl=https://github.com/Riverside-Software/sonar-openedge/releases/tag/V2.14.4
+2.14.4.downloadUrl=https://github.com/Riverside-Software/sonar-openedge/releases/download/V2.14.4/sonar-openedge-plugin-2.14.4.jar
+
 2.13.1.description=Initial Marketplace entry. Support for OpenEdge 12.3 and 12.4
-2.13.1.sqVersions=[7.6,8.2]
+2.13.1.sqVersions=[7.6,8.9]
```
ganncamp, line 16: Why are you suddenly upgrading the end of this range? There's already a replacement version out: 2.14.4. Further, the versions you're adding compatibility for are past EOL. The range on this old version needs to stay where it is, so users of the current LTS will adopt your latest version, rather than this old one.

openedge.properties
```diff
@@ -2,12 +2,18 @@ category=Languages
 description=Code Analyzer for OpenEdge
 homepageUrl=https://github.com/Riverside-Software/sonar-openedge
 archivedVersions=
-publicVersions=2.13.1
+publicVersions=2.13.1,2.14.4
```
ganncamp, line 5: You should move 2.13.1 to `archivedVersions`

ganncamp (changes requested):

openedge.properties
gquerret: Wouldn't that prevent SQ7.9 users from upgrading to the latest version of the plugin?
The range end was set to 9.1 because the plugin had been tested up to this version. My understanding was that plugins should be tested with the latest releases before upgrading this version number, but that may create a lot of PRs in this repo.  I've updated this PR to use LATEST.

openedge.properties
gquerret: Done

openedge.properties
gquerret: I was an error on my side, I thought those numbers reflected the minimum version of each major release, i.e. 7.6 and greater for 7.x branch, and 8.2 and greater for 8.x branch.
But as 2.13.1 is now an old version, this is not really important anymore

openedge.properties
ganncamp: 7.9 is fully past EOL. We do not support it any more. any where. That means here too. 7.9 users don't get plugin upgrades. They get to upgrade SonarQube itself.

openedge.properties
ganncamp: Please respond in the thread so I can refind it easily. :-)

gquerret:
> Please respond in the thread so I can refind it easily. :-)

PR has been updated. Response directly in the thread :-)

ganncamp:
I meant in the community thread since I need to respond to it publicly and that thread links here.

openedge.properties
```diff
@@ -2,10 +2,17 @@ category=Languages
...
 archivedVersions=
-publicVersions=2.13.1
+publicVersions=2.14.4
+archivedVersions=2.13.1
 defaults.mavenGroupId=eu.rssw.sonar.openedge
 defaults.mavenArtifactId=sonar-openedge-plugin
 
+2.14.4.description=New expression engine in ABL Parser
+2.14.4.sqVersions=[8.9,LATEST]
+2.14.4.date=2021-10-04
```
ganncamp, line 12:
```suggestion
2.14.4.date=2021-11-23
```

ganncamp (approved):

PR #216, opened 2021-05-21
SCM TFVC 2.3.1

lg2de:


scmtfvc.properties
```diff
@@ -2,20 +2,26 @@ category=Integration
...
 
+2.3.1.description=fix regression
+2.3.1.sqVersions=[7.9,LATEST]
+2.3.1.downloadUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/download/2.3.1/sonar-scm-tfvc-plugin-2.3.1.115.jar
+2.3.1.date=2021-05-21
+2.3.1.changelogUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/tag/2.3.1
+
 2.3.description=embed TFS API
-2.3.sqVersions=[7.9,LATEST]
+2.3.sqVersions=[7.9,8.8]
```
ganncamp, line 18: This really needs to end at 8.9.* or users on the latest version won't be prompted to upgrade (yes that still happens in CE)

ganncamp (changes requested):

scmtfvc.properties
lg2de: done

ganncamp (approved):

PR #217, opened 2021-05-25
sonar-clover 4.1

Tony-Proum:


christophelevis (changes requested):
Please start the compatibility range with 7.9. We dropped the compatibility with older versions.

christophelevis (approved):

PR #219, opened 2021-05-28
sonar-apigee 3.0.0

ntiss:
Declare release 3.0.0 for SQ 8.8+
Keep 2.1.1 for older SQ version
End of life of 2.0.0

### DIFF
--- a/apigee.properties
+++ b/apigee.properties
@@ -1,20 +1,21 @@
 category=External Analyzers
 description=Adds XML rules test Apigee apiproxies.
 homepageUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin
-archivedVersions=2.0.0
-publicVersions=2.1.0
+archivedVersions=2.1.1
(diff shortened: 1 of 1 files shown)


apigee.properties
```diff
@@ -1,20 +1,21 @@
 category=External Analyzers
 description=Adds XML rules test Apigee apiproxies.
 homepageUrl=https://github.com/CreditMutuelArkea/sonar-apigee-plugin
-archivedVersions=2.0.0
-publicVersions=2.1.0
+archivedVersions=2.1.1
+publicVersions=3.0.0
```
agigleux, line 5: You should put in the "publicVersions", all versions that can be still used and here the 2.1.1 is still useful for SQ 7.9-8.7.* users. So I think here, you should put:

```
archivedVersions=2.0.0
publicVersions=2.1.1,3.0.0
```

agigleux (changes requested):

PR #221, opened 2021-05-30
New Sonar-Groovy release 1.7

TobiX:
Announcement: https://community.sonarsource.com/t/new-release-sonar-groovy-plugin-1-7/44007

### DIFF
--- a/groovy.properties
+++ b/groovy.properties
@@ -2,15 +2,22 @@ category=Languages
 description=Enables scanning of Groovy source files
 homepageUrl=https://github.com/Inform-Software/sonar-groovy
 issueTrackerUrl=https://github.com/Inform-Software/sonar-groovy/issues
-publicVersions=1.6
+publicVersions=1.7
 privateVersions=
-archivedVersions=
+archivedVersions=1.6
 
 defaults.mavenGroupId=org.sonarsource.sonar-groovy-plugin
 defaults.mavenArtifactId=sonar-groovy-plugin
 
 1.6.description=Support for modern SonarQube versions, dependency updates
-1.6.sqVersions=[7.9,LATEST]
+1.6.sqVersions=[7.9,8.9.*]
 1.6.date=2019-07-04
 1.6.downloadUrl=https://github.com/Inform-Software/sonar-groovy/releases/download/1.6/sonar-groovy-plugin-1.6.jar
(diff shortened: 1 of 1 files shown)


groovy.properties
```diff
@@ -2,15 +2,22 @@ category=Languages
...
 
 1.6.description=Support for modern SonarQube versions, dependency updates
-1.6.sqVersions=[7.9,LATEST]
+1.6.sqVersions=[7.9,8.9.*]
 1.6.date=2019-07-04
 1.6.downloadUrl=https://github.com/Inform-Software/sonar-groovy/releases/download/1.6/sonar-groovy-plugin-1.6.jar
 1.6.changelogUrl=https://github.com/Inform-Software/sonar-groovy/blob/master/CHANGELOG.md#16---2019-07-04
+
+1.7.description=Minor dependency updates, deprecate binary JaCoCo support
+1.7.sqVersions=[7.9,LATEST]
```
ganncamp, line 19: this end range should be updated to the current version: 8.9

groovy.properties
```diff
@@ -2,15 +2,22 @@ category=Languages
...
 privateVersions=
-archivedVersions=
+archivedVersions=1.6
 
 defaults.mavenGroupId=org.sonarsource.sonar-groovy-plugin
 defaults.mavenArtifactId=sonar-groovy-plugin
 
 1.6.description=Support for modern SonarQube versions, dependency updates
-1.6.sqVersions=[7.9,LATEST]
+1.6.sqVersions=[7.9,8.9.*]
```
ganncamp, line 13: This should end with `LATEST`... unless you really mean to have this version marked incompatible when 9.0 comes out?

ganncamp (changes requested):

TobiX:
I'm a bit confused: I thought I would set the compatibility end range of the newest release to LATEST, so it will be automatically compatible with future versions and set the end rage of the old (now deprecated) version to the current release, so users are urged to update when installing a newer version of SonarQube. Is that not how it works?

ganncamp (approved):
Sorry. _I_ was confused. Most people add the new version at the top of the list & I didn't catch that you had done the opposite. As they say: haste makes waste.

TobiX:
No problem, thanks for merging!

PR #228, opened 2021-07-14
Update aemrules.properties

colin-mueller-sonarsource:
This plugin is not compatible with 8.9 based on Community and Support requests, as well as this issue on the repository:

https://github.com/wttech/AEM-Rules-for-SonarQube/issues/213


ganncamp (approved):
It seems really odd to me that an incompatibility would _start_ at 8.9 (surely it started earlier). But okay

PR #230, opened 2021-07-22
Added jQAssistant SonarQube Plugin 1.10.0

DirkMahler:


jqassistant.properties
```diff
@@ -1,15 +1,21 @@
...
+1.10.0.description=Upgraded compatibility to SonarQube 8.9 LTS and added configurable issue types
+1.10.0.date=2021-07-22
+1.10.0.sqVersions=[8.9,LATEST]
+1.10.0.changelogUrl=https://github.com/jqassistant-contrib/sonar-jqassistant-plugin/blob/master/release-notes.adoc#1100
+1.10.0.downloadUrl=https://search.maven.org/remotecontent?filepath=org/jqassistant/contrib/sonarqube/sonar-jqassistant-plugin/1.10.0/sonar-jqassistant-plugin-1.10.0.jar
+
 1.9.0.description=Create ad-hoc rules and external issues instead of generic concept/constraint issues where possible and detect reports from parent modules
 1.9.0.date=2020-10-13
-1.9.0.sqVersions=[7.9,LATEST]
+1.9.0.sqVersions=[7.9,8.5]
```
ganncamp, line 18: Unless you specifically know that this version is incompatible past 8.5, this range should end at 9.0. Otherwise your users on versions past 8.5 won't be prompted to upgrade to your latest version.

ganncamp (changes requested):

jqassistant.properties
DirkMahler: Thanks for the feedback, just fixed it.

ganncamp (approved):

PR #231, opened 2021-07-25
Released yaml plugin version 1.6.0

sbaudoin:


yaml.properties
```diff
@@ -1,12 +1,18 @@
...
-archivedVersions=1.4.3,1.5.0,1.5.1
-publicVersions=1.5.2
+archivedVersions=1.4.3,1.5.0,1.5.1,1.5.2
+publicVersions=1.6.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
+1.6.0.description=Allows rule customization from a local .yamllint file and fixes some documentation inconsistencies
+1.6.0.sqVersions=[7.9,8.9.*]
```
ganncamp, line 11: Do you intend this version not to be compatible with 9.0 & later? I suggest you change the end value to `LATEST`.

ganncamp (changes requested):

yaml.properties
sbaudoin: Hello,
Just before creating this PR (version 1.6.0 was just released...), I saw that one of your colleague created a PR for compatibility of the YAML plugin with SQ 9.1 (not released yet but this was not 9.0, though). So I decided to released anyway the plugin but for 7.* and 8.* SQ releases, waiting for a feedback of your colleague to state about the actual compatibility status with SQ 9. Now I have the answer: my plugin is compatible with SQ 9.0 but not with SQ 9.1.
So, back to this PR: I have not tested it with SQ 9.0 so I cannot say if it is compatible with SQ 9.0. So I suggest that I run my tests with SQ 9.0 and if this is OK and if you agree, I'll update this PR only with 9.0 as the version for SQ 9, not 9.* or LATEST, waiting for SQ 9.1 to be available to release a new version of the plugin including your colleague's PR and this time for 7.9, 8.9.* and LATEST.

ganncamp: Fine for me. We can even release now with 8.9.* if you like. This is totally up to you (until and unless we discover an incompatibility).

sbaudoin: I'm disappointed: I'm facing SONAR-15167 while running my integration tests on Travis CI. I fear I won't be able to play with SQ 9 on Travis until they upgrade Docker. So in order to prevent waiting for too long, I'd prefer push this 1.6.0 release for SQ 7 and 8 only and I'll request an update when I am able to fully test on SQ 9. So far this plugin cannot be said to be compatible for SQ 9.

sbaudoin: Seems that I managed to have Travis to run my tests... So if you may wait a bit more ^_^

sbaudoin: So the tests pass with SQ 9.0: I add this version.

yaml.properties
```diff
@@ -1,12 +1,18 @@
...
-archivedVersions=1.4.3,1.5.0,1.5.1
-publicVersions=1.5.2
+archivedVersions=1.4.3,1.5.0,1.5.1,1.5.2
+publicVersions=1.6.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
+1.6.0.description=Allows rule customization from a local .yamllint file and fixes some documentation inconsistencies
+1.6.0.sqVersions=[7.9,8.9.*,9.0]
```
ganncamp, line 11: The brackets (`[...]`) format describes a range, so this value isn't valid. Your choices are (note that a point version of 9.0 has been released):
* `[7.9,8.9.*],9.0.*`
* `[7.9,9.0.*]`

I suggest the latter

ganncamp (changes requested):

yaml.properties
sbaudoin: Ok, thanks for the details, I go for the latter as well. New commit to come.

ganncamp (approved):

PR #245, opened 2021-09-29
Add Sonar CNES Report version 4.0.0

Sancretor:
# Changelog
* SonarQube 8.9 LTS compatibility
* Add new metrics data in the reports (Quality Gate status, detailed technical debt, adherence to coding standard, etc.)
* Add graphs in report (evolution of the number of issues, evolution of technical debt ratio)
* Various bugfixes
 
And more details here : https://github.com/cnescatlab/sonar-cnes-report/releases/tag/4.0.0

*Compatibility only includes 8.9 LTS because we rely a lot on SQ APIs. 9.x versions may change them, and we don't have enough time to maintain the plugin for every intermediate SQ release before the next LTS.*

### DIFF
--- a/cnesreport.properties
+++ b/cnesreport.properties
@@ -7,7 +7,7 @@ publicVersions=4.0.0
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
-4.0.0.description=Add compatibility with SonarQube 8.9 LTS, enhance reports with new metrics quality gate status, adherence to coding standard, etc.) and graphs (number of issues, technical debt ratio), and various bugfixes. status, adherence to coding standard, etc.) and graphs (number of issues, technical debt ratio), and va
+4.0.0.description=Add compatibility with SonarQube 8.9 LTS, enhance reports with new metrics (quality gate status, adherence to coding standard, etc.) and graphs (number of issues, technical debt ratio), and various bugfixes.
 4.0.0.date=2021-09-23
 4.0.0.sqVersions=[8.9]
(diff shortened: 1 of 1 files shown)


cnesreport.properties
```diff
@@ -1,17 +1,24 @@
...
-publicVersions=3.3.1
+archivedVersions=3.1.0,3.2.2,3.3.1
+publicVersions=4.0.0
 
 defaults.mavenGroupId=fr.cnes.sonarqube.plugins
 defaults.mavenArtifactId=cnesreport
 
+4.0.0.description=Add compatibility with SonarQube 8.9 LTS, enhance reports with new metrics (quality gate status, adherence to coding standard, etc.) and graphs (number of issues, technical debt ratio), and various bugfixes.
+4.0.0.date=2021-09-23
+4.0.0.sqVersions=[8.9]
```
ganncamp, line 12: If you really want to declare compatibility only with a single version then you don't need range syntax (`[]`), altho it doesn't break anything. But do you really mean to say this isn't compatible with 9.0, 9.1 &etc.?

ganncamp (changes requested):

cnesreport.properties
Sancretor: We did not work with SQ 9.0/9.1, only with LTS versions for the sake of simplicity.
As stated in the description, we rely on SQ API and they may evolve on non-LTS releases.
It may actually work for 9.0 or 9.1, but by indicating so, people will ask us for a quick 9.2 compatibility, and so on...
I'm afraid we won't have enough free time to update the plugin on a regular basis.

ganncamp (approved):

PR #252, opened 2021-11-14
Cloudformation plugin 2.1.8

pethers:
Add latest release

### DIFF
--- a/cloudformation.properties
+++ b/cloudformation.properties
@@ -2,17 +2,17 @@ name=Cloudformation
 category=External Analysers
 descripton=Cloudformation rules for json and yaml, integrates cfn-nag reports.
 homepageUrl=https://github.com/Hack23/sonar-cloudformation-plugin
-archivedVersions=2.0.1,2.0.2,2.0.3,2.0.4,2.0.5,2.0.7,2.0.8,2.0.9,2.1.18
-publicVersions=2.1.18
+archivedVersions=2.0.1,2.0.2,2.0.3,2.0.4,2.0.5,2.0.7,2.0.8,2.0.9,2.0.10
+publicVersions=2.1.8
 
 defaults.mavenGroupId=com.hack23.sonar
 defaults.mavenArtifactId=sonar-cloudformation-plugin
 
-2.1.18.description=Adds support for checkov. Include rules for terraform, kubernetes and azure templates.
-2.1.18.date=2021-11-09
-2.1.18.sqVersions=[7.9,8.9.*]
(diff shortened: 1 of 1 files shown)


cloudformation.properties
```diff
@@ -2,12 +2,18 @@ name=Cloudformation
...
-publicVersions=2.0.10
+archivedVersions=2.0.1,2.0.2,2.0.3,2.0.4,2.0.5,2.0.7,2.0.8,2.0.9,2.1.18
+publicVersions=2.1.18
 
 defaults.mavenGroupId=com.hack23.sonar
 defaults.mavenArtifactId=sonar-cloudformation-plugin
 
+2.1.18.description=Adds support for checkov. Include rules for terraform, kubernetes and azure templates.
+2.1.18.date=2021-11-09
+2.1.18.sqVersions=[7.9,8.9.*]
```
ganncamp, line 13: All support for 7.9.* is officially over. This range needs to start at 8.9. And do you really mean to _end_ it at 8.9 & not at least support 9.1?
Also, <checks email archives>, if you want to continue supporting Checkov, CNF_NAG &etc reports if there's no conflict with a double-report of metrics. I guess that would be a different version?

cloudformation.properties
```diff
@@ -2,12 +2,18 @@ name=Cloudformation
...
+2.1.18.changelogUrl=https://github.com/Hack23/sonar-cloudformation-plugin/releases/tag/sonar-cloudformation-plugin-2.1.18
```
ganncamp, line 14: 404 for me

cloudformation.properties
```diff
@@ -2,12 +2,18 @@ name=Cloudformation
...
+2.1.18.downloadUrl=https://oss.sonatype.org/service/local/repositories/releases/content/com/hack23/sonar/sonar-cloudformation-plugin/2.1.18/sonar-cloudformation-plugin-2.1.18.jar
```
ganncamp, line 15: also 404

ganncamp (changes requested):

cloudformation.properties
pethers: Only tested/use it on 8.9 and old LTS. Might work on other versions but need to setup some integration test before I want to say it is supported.

cloudformation.properties
pethers: Release I made an error, the release was 2.1.8 and not 2.1.18 so did some cut and paste errors. But corrected but links now.

ganncamp (approved):

PR #253, opened 2021-11-17
Added release 2.5.0 for shellcheck

sbaudoin:


shellcheck.properties
```diff
@@ -1,14 +1,20 @@
...
-archivedVersions=2.2.0,2.2.1,2.3.0
-publicVersions=2.4.0
+archivedVersions=2.2.0,2.2.1,2.3.0,2.4.0
+publicVersions=2.5.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-shellcheck-plugin
 
+2.5.0.description=Brings 54 new rules, fixes some minor issues
+2.5.0.sqVersions=[7.9,LATEST]
```
ganncamp, line 11: all support for 7.9 is officially over. Please update the range start to the current LTS; 8.9

shellcheck.properties
```diff
@@ -1,14 +1,20 @@
...
+2.5.0.date=2021-11-17
+2.5.0.changelogUrl=https://github.com/sbaudoin/sonar-shellcheck/releases/tag/v2.5.0
+2.5.0.downloadUrl=https://github.com/sbaudoin/sonar-shellcheck/releases/download/v2.5.0/sonar-shellcheck-plugin-2.5.0.jar
+
 2.4.0.description=Fixes issues (including a security issue), brings 75 new rules, now highlights Shell code even if the plugin has been disabled
-2.4.0.sqVersions=[7.9,LATEST]
+2.4.0.sqVersions=[7.9,8.9.*]
```
ganncamp, line 17: This range should end with the current version, otherwise your users on versions later than 8.9.* won't be prompted in-app to upgrade

ganncamp (changes requested):

shellcheck.properties
sbaudoin: Sorry I don't get what range should be set: 2.4 should work on SQ 9+ but I did not tested it. Instead people willing to use SQ 9 should use the 2.5.0 version.

ganncamp: This range previously ended with `LATEST`. That means users on all SQ new versions were offered this plugin version through the Marketplace. So people on 9.2 may have this installed. When you end the range with 8.9, then you're telling the Marketplace that people on 8.9 should be prompted to upgrade... but people on 9.1 and 9.2 _should not_. So end this range with the current version so that _all_ your users will be prompted to upgrade.

ganncamp (changes requested):

shellcheck.properties
```diff
@@ -1,14 +1,20 @@
...
 
+2.5.0.description=Brings 54 new rules, fixes some minor issues
+2.5.0.sqVersions=[8.9,LATEST]
+2.5.0.date=2021-11-17
+2.5.0.changelogUrl=https://github.com/sbaudoin/sonar-shellcheck/releases/tag/v2.5.0
+2.5.0.downloadUrl=https://github.com/sbaudoin/sonar-shellcheck/releases/download/v2.5.0/sonar-shellcheck-plugin-2.5.0.jar
+
 2.4.0.description=Fixes issues (including a security issue), brings 75 new rules, now highlights Shell code even if the plugin has been disabled
-2.4.0.sqVersions=[7.9,LATEST]
+2.4.0.sqVersions=[7.9,9.2]
```
ganncamp, line 17:
```suggestion
2.4.0.sqVersions=[7.9,9.2.*]
```

ganncamp (changes requested):

ganncamp (approved):

PR #260, opened 2021-12-06
Sonar rust upgrade to v0.0.7

elegoff:
Fixed requested changes on sqVersions


communityrust.properties
```diff
@@ -1,13 +1,20 @@
 category=Languages
 description=Code Analyzer for Rust
 homepageUrl=https://github.com/elegoff/sonar-rust
-publicVersions=0.0.6
+
```
ganncamp, line 4: There should be an `archivedVersions` list, and 0.0.6 should be in it.

ganncamp (changes requested):

ganncamp (approved):

PR #263, opened 2021-12-28
OIDC Auth 2.1.1

tjuerge:
Updated OpenID Connect Authentication Plugin to v2.1.1:

* fixes some issues with the newly added auto-login servlet filter

### DIFF
--- a/authoidc.properties
+++ b/authoidc.properties
@@ -8,7 +8,7 @@ defaults.mavenGroupId=org.vaulttec.sonarqube.auth.oidc
 defaults.mavenArtifactId=sonar-auth-oidc-plugin
 
 2.1.1.description=Fixes some issues with the newly added auto-login servlet filter
-2.1.1.sqVersions=[7.9,LATEST]
+2.1.1.sqVersions=[8.9,LATEST]
 2.1.1.date=2021-10-24
 2.1.1.changelogUrl=https://github.com/vaulttec/sonar-auth-oidc/releases/tag/v2.1.1
 2.1.1.downloadUrl=https://github.com/vaulttec/sonar-auth-oidc/releases/download/v2.1.1/sonar-auth-oidc-plugin-2.1.1.jar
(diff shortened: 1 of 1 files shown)


authoidc.properties
```diff
@@ -1,12 +1,18 @@
...
-archivedVersions=1.0.4,1.1.0,2.0.0
-publicVersions=2.1.0
+archivedVersions=1.0.4,1.1.0,2.0.0,2.1.0
+publicVersions=2.1.1
 
 defaults.mavenGroupId=org.vaulttec.sonarqube.auth.oidc
 defaults.mavenArtifactId=sonar-auth-oidc-plugin
 
+2.1.1.description=Fixes some issues with the newly added auto-login servlet filter
+2.1.1.sqVersions=[7.9,LATEST]
```
ganncamp, line 11: This range needs to start with 8.9. 
7.9 is fully past EOL and sunsetted now.

authoidc.properties
```diff
@@ -1,12 +1,18 @@
...
+2.1.1.date=2021-10-24
```
ganncamp, line 12: This date doesn't look right to me. You posted the request in the community on 28 December.

ganncamp (changes requested):

authoidc.properties
tjuerge: This plugin is [compiled and tested against SQ 7.4](https://github.com/vaulttec/sonar-auth-oidc/blob/master/pom.xml#L26). But as requested I'll limit this plugin's marketplace entry to SQ 8.9 and newer.

authoidc.properties
tjuerge: Yes, it took me 2 month (after the release of this plugin version on 2021-10-24) to prepare the steps for the marketplace entry :)
Screenshot 2022-01-06 at 23 15 19

ganncamp: This date should be (approximately) when the plugin became available in the Marketplace.

ganncamp (changes requested):

authoidc.properties
tjuerge: Ok, I'll fix this...

ganncamp (approved):

PR #265, opened 2022-01-15
Update sonar-detekt to 2.5.0

cortinico:
We released sonar-detekt version 2.5.0, see: https://github.com/detekt/sonar-kotlin/releases/tag/2.5.0

which now bundles Detekt 1.19.0.

### DIFF
--- a/detekt.properties
+++ b/detekt.properties
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Provide detekt rules for Kotlin projects
 homepageUrl=https://github.com/detekt/sonar-kotlin
-archivedVersions=2.0.0,2.2.0,2.3.0
-publicVersions=2.1.0,2.4.0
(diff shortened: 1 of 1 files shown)


detekt.properties
```diff
@@ -1,14 +1,20 @@
 category=External Analysers
 description=Provide detekt rules for Kotlin projects
 homepageUrl=https://github.com/detekt/sonar-kotlin
-archivedVersions=2.0.0,2.2.0,2.3.0
-publicVersions=2.1.0,2.4.0
+archivedVersions=2.0.0,2.2.0,2.3.0,2.4.0
+publicVersions=2.1.0,2.5.0
```
ganncamp, line 5: It's time to archive 2.1.0. It only supports 7.9, which is EOL.

ganncamp (changes requested):

ganncamp (approved):

PR #287, opened 2022-04-12
UPC-116 Remove dependency on legacy plugins for Compatibility Matrix Generation

pierre-guillot-sonarsource:
relates to https://github.com/SonarSource/sonar-update-center/pull/70, which needs to be merged at the same time.


aurelien-poscia-sonarsource (approved):
LGTM, the only question is: is it worth to have N commits instead of 1?

leo-geoffroy-sonarsource (approved):

PR #312, opened 2022-05-11
new sonar-pmd-plugin: 3.4.0

jborgers:
We released a new version of the sonar-pmd-plugin to support Java 18 and Sonar 8.9/9. Please make it available in the marketplace. Thanks!

### DIFF
--- a/pmd.properties
+++ b/pmd.properties
@@ -1,15 +1,15 @@
 category=External Analysers
 description=Provide PMD rules to analyse Java projects
-archivedVersions=
-#publicVersions=3.3.1
+archivedVersions=3.3.1
+publicVersions=3.4.0
 
 defaults.mavenGroupId=org.sonarsource.pmd
 defaults.mavenArtifactId=sonar-pmd-plugin
 
-3.3.1.description=Upgrade to PMD 6.30 and Java 15 support
-#3.3.1.sqVersions=[8.9,LATEST]
-3.3.1.date=2021-02-06
-3.3.1.changelogUrl=https://github.com/jensgerdes/sonar-pmd/releases/tag/3.3.1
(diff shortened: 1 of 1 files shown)


ganncamp (changes requested):
You need to retain the 3.3.1 information and _add_ a 3.4.0 block (my preference is on top of the list, but it doesn't really matter). Take a look at https://github.com/SonarSource/sonar-update-center-properties/blob/master/dependencycheck.properties as an example

jborgers:
3.3.1 information retained (added)

pmd.properties
```diff
@@ -1,15 +1,21 @@
...
 
+3.4.0.description=Upgrade to PMD 6.45 and Java 18 support
+3.4.0.sqVersions=[8.9,LATEST]
+3.4.0.date=2022-05-11
+3.4.0.changelogUrl=https://github.com/jborgers/sonar-pmd/releases/tag/3.4.0
+3.4.0..downloadUrl=https://github.com/jborgers/sonar-pmd/releases/download/3.4.0/sonar-pmd-plugin-3.4.0.jar
+
 3.3.1.description=Upgrade to PMD 6.30 and Java 15 support
-#3.3.1.sqVersions=[8.9,LATEST]
+3.3.1.sqVersions=[8.9,LATEST]
```
ganncamp, line 16: Only one version can be compatible with `LATEST` at a time. 
When archiving the previous version, you need to update the compatibility range to end with the current version of SonarQube. So 9.4 in this case.

ganncamp (changes requested):

pmd.properties
ganncamp: Only one version can be compatible with `LATEST` at a time. 
When archiving the previous version, you need to update the compatibility range to end with the current version of SonarQube. So 9.4 in this case.

jborgers:
3.3.1 compatibility range updated

ganncamp (approved):

PR #334, opened 2022-07-13
Added jQAssistant SonarQube Plugin 1.12.0

DirkMahler:


jqassistant.properties
```diff
@@ -1,15 +1,21 @@
 category=External Analyzers
 description=Analyze issues reported by jQAssistant.
 homepageUrl=https://github.com/jqassistant-contrib/sonar-jqassistant-plugin
-archivedVersions=1.9.0,1.10.0
-publicVersions=1.11.0
+archivedVersions=1.7.3,1.8.1,1.8.2,1.9.0,1.10.0,1.11.0
```
ganncamp, line 4: You're adding a lot of version numbers back into this list when the metadata for those versions no longer exists in the file. I'm pretty sure the job will fail with these missing versions. They were removed from the file because they only supported versions that are past EOL. Please revert the changes an only add the newly-archived version.

ganncamp (changes requested):

jqassistant.properties
DirkMahler: The outdated versions are removed now.

ganncamp (approved):

PR #341, opened 2022-08-15
Update Azure DevOps extension

christophe-havard-sonarsource:


mickael-caro-sonarsource:
@christophe-havard-sonarsource I would suggest putting the "real" release date of it, which was the 9th of August.

Let's also figure out first why the changelog URL doesn't work. 

Last, the title of this PR is not correct, Update is for the Azure DevOps extension, not the Scanner for .NET itself. 

Thanks!

mickael-caro-sonarsource (approved):
LGTM

PR #343, opened 2022-08-19
Sonar rust

elegoff:


communityrust.properties
```diff
@@ -1,14 +1,23 @@
...
+
+0.1.1.description=Rust Grammar maintenance  
+0.1.1.sqVersions=[8.9,LATEST]
+0.1.1.date=2022-08-19
+0.1.1.changelogUrl=https://github.com/elegoff/sonar-rust/releases/tag/v0.1.1
+0.1.1.downloadUrl=https://github.com/elegoff/sonar-rust/releases/download/v0.1.1/community-rust-plugin-0.1.1.jar
+
 0.1.0.description=Better grammar parsing 
-0.1.0.sqVersions=[8.9,LATEST]
+0.1.0.sqVersions=[8.9,9.5]
```
ganncamp, line 19: 9.6 was announced on Monday. Some of your users may have already upgraded, so this range should end with 9.6. Otherwise, they won't be prompted to upgrade.
Don't worry, this version won't be offered to 9.6 users for new install.

ganncamp (changes requested):

communityrust.properties
ganncamp: This range still ends with 9.5...?

ganncamp (changes requested):

ganncamp (approved):

PR #383, opened 2023-02-12
Release findbugs 4.2.3

gtoison:
Hello,

We are releasing version 4.2.3 of the FindBugs plugin:
Download link: https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.2.3/sonar-findbugs-plugin-4.2.3.jar
Release note: https://github.com/spotbugs/sonar-findbugs/releases/tag/4.2.3
SonarCloud: https://sonarcloud.io/summary/new_code?id=com.github.spotbugs%3Asonar-findbugs-plugin

Thank you!


findbugs.properties
```diff
@@ -45,7 +45,12 @@ defaults.mavenArtifactId = sonar-findbugs-plugin
 4.2.1.changelogUrl = https://github.com/spotbugs/sonar-findbugs/releases/tag/4.2.1
 4.2.1.downloadUrl = https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.2.1/sonar-findbugs-plugin-4.2.1.jar
 4.2.2.description = Use SpotBugs 4.7.3, sb-contrib 7.4.7, and findsecbugs 1.12.0
-4.2.2.sqVersions = [8.9,LATEST]
+4.2.2.sqVersions = [8.9,9.9.*]
 4.2.2.date = 2022-10-20
 4.2.2.changelogUrl = https://github.com/spotbugs/sonar-findbugs/releases/tag/4.2.2
 4.2.2.downloadUrl = https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.2.2/sonar-findbugs-plugin-4.2.2.jar
+4.2.3.description = Use SpotBugs 4.7.3, sb-contrib 7.4.7, and findsecbugs 1.12.0
+4.2.3.sqVersions = [8.9,LATEST]
```
ganncamp, line 53: Should this really be listed as compatible with the _previous_ LTS? Especially when [issue 18](https://github.com/spotbugs/sonar-findbugs/issues/18#issuecomment-1427011510), which is fixed in this version, requires SonarQube 9.8+?  

Since 9.9 is the current LTS, it would be easiest all around IMO to start compatibility with 9.9

findbugs.properties
```diff
@@ -45,7 +45,12 @@ defaults.mavenArtifactId = sonar-findbugs-plugin
...
+4.2.3.date = 2023-02-12
+4.2.3.changelogUrl = https://github.com/spotbugs/sonar-findbugs/releases/tag/4.2.3
```
ganncamp, line 55: Normally I don't know enough to comment on the contents of a changelog, but...
I would have expected to see [#18](This should be fixed in version 4.2.3 (just released)
Note that the fix needs the SonarQube version to be >= 9.8) listed in the changelog and I don't...?

ganncamp (changes requested):

findbugs.properties
gtoison: This version fixes another longstanding issue (a file leak), it is as compatible with 8.9 as the previous versions were. 
Starting the compatibility with 9.9 would prevent people still on 8.9 from benefiting of that fix. Actually the plugin is still tested on 7.9 and works fine too.

The change related to 9.8+ is compatible with earlier versions, it's just that it won't do anything before 9.8

findbugs.properties
gtoison: I'm not sure if it's a valid justification but historically the "description" was always listing the SpotBugs (& plugins) versions, without any other details.
If the "description" is actually meant to be a changelog, I suppose I can add more info here. Let me know what you think!

ganncamp: I'm for whatever gives users access to a comprehensive list of changes. I only question this because what I end up at (https://github.com/spotbugs/sonar-findbugs/compare/4.2.2...4.2.3) doesn't include something that I was pointed to this version for...?

ganncamp: The changelog referenced here ends with
> Full Changelog: https://github.com/spotbugs/sonar-findbugs/compare/4.2.2...4.2.3

ganncamp (approved):

gtoison:
Thank you!

PR #403, opened 2023-03-31
Unlist the CaYC plugin  temporarily

vivek-reghunath-sonarsource:
Unlisting the CaYC plugin until the limitation on installation from the marketplace is addressed.


christophelevis (approved):
👍

PR #413, opened 2023-05-16
Update sonar-dependency-check-plugin to 4.0.0

Reamer:
We've released sonar-dependency-check-plugin version 4.0.0, please add it to the marketplace.
Thanks in advance!

SonarCloud-Dashboard: https://sonarcloud.io/dashboard?id=dependency-check_dependency-check-sonar-plugin

### DIFF
--- a/dependencycheck.properties
+++ b/dependencycheck.properties
@@ -1,13 +1,19 @@
 category=Integration
 description=Integrates Dependency-Check reports into SonarQube
 homepageUrl=https://github.com/dependency-check/dependency-check-sonar-plugin
-archivedVersions=2.0.7,2.0.8,3.0.0,3.0.1
-publicVersions=3.1.0
+archivedVersions=2.0.7,2.0.8,3.0.0,3.0.1,3.1.0
+publicVersions=4.0.0
 defaults.mavenGroupId=org.sonarsource.owasp
 defaults.mavenArtifactId=sonar-dependency-check-plugin
 
+4.0.0.description=Support SonarQube 10
+4.0.0.sqVersions=[9.9.1, LATEST]
+4.0.0.date=2023-05-16
+4.0.0.changelogUrl=https://github.com/dependency-check/dependency-check-sonar-plugin/releases/tag/4.0.0
+4.0.0.downloadUrl=https://github.com/dependency-check/dependency-check-sonar-plugin/releases/download/4.0.0/sonar-dependency-check-plugin-4.0.0.jar
+
 3.1.0.description=Support dependency-check 8.0.0
-3.1.0.sqVersions=[8.9.7, LATEST]
+3.1.0.sqVersions=[8.9.7, 9.9]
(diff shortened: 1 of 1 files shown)


dependencycheck.properties
```diff
@@ -1,13 +1,19 @@
...
 
+4.0.0.description=Support SonarQube 10
+4.0.0.sqVersions=[9.9.1, LATEST]
+4.0.0.date=2023-05-16
+4.0.0.changelogUrl=https://github.com/dependency-check/dependency-check-sonar-plugin/releases/tag/4.0.0
+4.0.0.downloadUrl=https://github.com/dependency-check/dependency-check-sonar-plugin/releases/download/4.0.0/sonar-dependency-check-plugin-4.0.0.jar
+
 3.1.0.description=Support dependency-check 8.0.0
-3.1.0.sqVersions=[8.9.7, LATEST]
+3.1.0.sqVersions=[8.9.7, 9.9]
```
ganncamp, line 16: The range end needs to be set to the current version. Otherwise your users on 10.0 won't be prompted to upgrade.

dependencycheck.properties
```diff
@@ -1,13 +1,19 @@
...
 homepageUrl=https://github.com/dependency-check/dependency-check-sonar-plugin
-archivedVersions=2.0.7,2.0.8,3.0.0,3.0.1
-publicVersions=3.1.0
+archivedVersions=2.0.7,2.0.8,3.0.0,3.0.1,3.1.0
+publicVersions=4.0.0
 defaults.mavenGroupId=org.sonarsource.owasp
 defaults.mavenArtifactId=sonar-dependency-check-plugin
 
+4.0.0.description=Support SonarQube 10
+4.0.0.sqVersions=[9.9.1, LATEST]
```
ganncamp, line 10: Do you really mean to exclude users on 9.9(.0)? Unless that's a deliberate exclusion, this should start at 9.9

ganncamp (changes requested):

dependencycheck.properties
ganncamp: Okay. Then this should end with `9.9.*` or `9.9.1` explicitly since there is a point version out.

ganncamp (changes requested):

dependencycheck.properties
ganncamp: You're right that the latest point version is considered the LTS. And unfortunately, that doesn't mean that everyone upgrades to it immediately. By starting this at 9.9, people who haven't upgraded the server yet will still be prompted to upgrade their version of your plugin.

Reamer:
I have adjusted the versions. Thank you for your support.

ganncamp (approved):

Reamer:
TravisCI is failing. Do you have any idea why?

ganncamp (changes requested):

ganncamp (approved):

PR #414, opened 2023-05-26
REL-2664 Release Scanner for Gradle 4.1.0.3113

kaufco:


christophelevis (approved):
Thanks @kaufco. I made some updates and will make public.

PR #416, opened 2023-05-31
sonar-l10n-fr-plugin-9.9.1

jycr:
Hi,

sonar-l10n-fr-plugin-9.9.1 has been released.

Changes:

* Fixes misleading translations
* Improves "CI" translations
* Uses right apostrophe quotation mark

The Download link and release notes: https://github.com/jycr/sonar-l10n-fr/releases/tag/9.9.1

The SonarCloud: https://sonarcloud.io/project/overview?id=jycr_sonar-l10n-fr

Thank you

Regards

### DIFF
--- a/l10nfr.properties
+++ b/l10nfr.properties
@@ -1,8 +1,8 @@
 category=Localization
 description=Language Pack for French
 homepageUrl=https://github.com/SonarCommunity/sonar-l10n-fr
-publicVersions=1.15.1,9.8.1,9.9.0
-archivedVersions=9.8.0
+publicVersions=1.15.1,9.8.1,9.9.1
+archivedVersions=9.8.0,9.9.0
 
 defaults.mavenGroupId=org.sonarqube.l10n.fr
 defaults.mavenArtifactId=sonar-l10n-fr-plugin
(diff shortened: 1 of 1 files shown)


l10nfr.properties
```diff
@@ -27,7 +27,13 @@ defaults.mavenArtifactId=sonar-l10n-fr-plugin
 9.8.1.changelogUrl=https://github.com/jycr/sonar-l10n-fr/releases/tag/9.8.1
 
 9.9.0.description=Support SonarQube 9.9+
-9.9.0.sqVersions=[9.9,LATEST]
+9.9.0.sqVersions=9.9
```
ganncamp, line 30: You can't have two public versions with overlapping compatibility. The Marketplace has to know which plugin version to recommend for each SQ version.
I suggest you archive this one and set its compatibility to: `[9.9,9.9.*]` so that users on 9.9.0 will be prompted to upgrade

ganncamp (changes requested):

l10nfr.properties
jycr: Thanks for your advice

PR #449, opened 2023-10-02
Update scannermsbuild.properties for 5.14

tom-howlett-sonarsource:


scannermsbuild.properties
```diff
@@ -16,6 +16,15 @@ flavors.core20.label=.NET Core 2.1
 flavors.core31.label=.NET Core 3.1
 flavors.framework46.label=.NET Framework 4.6+
 
+5.14.description=Support SonarQube 10.3 API
```
christophelevis, line 19: I don't think this is a change related to 10.3. The drop should actually happen in 10.4 ([SONAR-19233](https://sonarsource.atlassian.net/browse/SONAR-19233)). Should we simply mention "Support for future versions of SonarQube"?

tom-howlett-sonarsource: I updated, I think its good to mention its 10.4 related

scannermsbuild.properties
```diff
@@ -16,6 +16,15 @@ flavors.core20.label=.NET Core 2.1
 flavors.core31.label=.NET Core 3.1
 flavors.framework46.label=.NET Framework 4.6+
 
+5.14.description=Support SonarQube 10.4 API changes
```
christophelevis, line 19: Should we use the word "Prepare..." since we are in advance?

christophelevis (approved):
LGTM. Just made a suggestion.

PR #453, opened 2023-10-23
Update yaml.properties

sbaudoin:
Added version 1.8.0
Project dashboard: https://sonarcloud.io/project/overview?id=sbaudoin_sonar-yaml

### DIFF
--- a/yaml.properties
+++ b/yaml.properties
@@ -1,15 +1,21 @@
 category=Languages
 description=Allows the analysis of YAML documents
 homepageUrl=https://github.com/sbaudoin/sonar-yaml
-archivedVersions=1.5.2,1.6.0
-publicVersions=1.7.0
+archivedVersions=1.5.2,1.6.0,1.7.0
+publicVersions=1.8.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
 
+1.8.0.description=Brings support for SonarQube 9.4+ and 10; adds a rule and improves some others; fixes some issues
+1.8.0.sqVersions=[8.9,LATEST]
+1.8.0.date=2023-10-23
+1.8.0.changelogUrl=https://github.com/sbaudoin/sonar-yaml/releases/tag/v1.8.0
+1.8.0.downloadUrl=https://github.com/sbaudoin/sonar-yaml/releases/download/v1.8.0/sonar-yaml-plugin-1.8.0.jar
+
 1.7.0.description=Allows rule customization from a local .yamllint file and fixes some documentation inconsistencies
-1.7.0.sqVersions=[8.9,LATEST]
+1.7.0.sqVersions=[8.9,9.3.*]
 1.7.0.date=2021-11-24
 1.7.0.changelogUrl=https://github.com/sbaudoin/sonar-yaml/releases/tag/v1.7.0
 1.7.0.downloadUrl=https://github.com/sbaudoin/sonar-yaml/releases/download/v1.7.0/sonar-yaml-plugin-1.7.0.jar


yaml.properties
```diff
@@ -1,15 +1,21 @@
...
 
+1.8.0.description=Brings support for SonarQube 9.4+ and 10; adds a rule and improves some others; fixes some issues
+1.8.0.sqVersions=[8.9,LATEST]
+1.8.0.date=2023-10-23
+1.8.0.changelogUrl=https://github.com/sbaudoin/sonar-yaml/releases/tag/v1.8.0
+1.8.0.downloadUrl=https://github.com/sbaudoin/sonar-yaml/releases/download/v1.8.0/sonar-yaml-plugin-1.8.0.jar
+
 1.7.0.description=Allows rule customization from a local .yamllint file and fixes some documentation inconsistencies
-1.7.0.sqVersions=[8.9,LATEST]
+1.7.0.sqVersions=[8.9,9.3.*]
```
ganncamp, line 18: Since the previous end value for this range was `LATEST`, the new value needs to be the current version. Otherwise, your users on 9.4+ won't be prompted to upgrade.

yaml.properties
```diff
@@ -1,15 +1,21 @@
...
-publicVersions=1.7.0
+archivedVersions=1.5.2,1.6.0,1.7.0
+publicVersions=1.8.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
 
+1.8.0.description=Brings support for SonarQube 9.4+ and 10; adds a rule and improves some others; fixes some issues
+1.8.0.sqVersions=[8.9,LATEST]
```
ganncamp, line 12: Anything older than 9.9 is officially EOL. This probably needs to start at 9.9

yaml.properties
```diff
@@ -1,15 +1,21 @@
...
-archivedVersions=1.5.2,1.6.0
-publicVersions=1.7.0
+archivedVersions=1.5.2,1.6.0,1.7.0
+publicVersions=1.8.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
 
+1.8.0.description=Brings support for SonarQube 9.4+ and 10; adds a rule and improves some others; fixes some issues
```
ganncamp, line 11: 9.4 is pretty old at this point. Could you maybe talk about 9.9 instead, please?

ganncamp (changes requested):

yaml.properties
sbaudoin: Why not. The thing is that 1.7 is known not to work with SQ 9.4 and beyond but version 1.8 works with SQ 9.4 and beyond. I just wanted to be as sharp as possible in the versions and allow the SQ 9.x users to be able to use the latest available version as it now works as of SQ 9.4.

Still want to mention 9.9 only?

yaml.properties
sbaudoin: The problem is that 1.7 does not work with SQ 9.4+

yaml.properties
```diff
@@ -1,15 +1,21 @@
...
-publicVersions=1.7.0
+archivedVersions=1.5.2,1.6.0,1.7.0
+publicVersions=1.8.0
 
 defaults.mavenGroupId=com.github.sbaudoin
 defaults.mavenArtifactId=sonar-yaml-plugin
 
 
+1.8.0.description=Brings support for SonarQube 9 LTS and 10; adds a rule and improves some others; fixes some issues
+1.8.0.sqVersions=[9.9.*,LATEST]
```
ganncamp, line 12: It turns out you can't use a wildcard at the start of a range. This needs to start with `9.9`

ganncamp (approved):

ganncamp (approved):

PR #457, opened 2023-11-15
YAML: new versions 1.8.1, 1.9.0 and 1.9.1

sbaudoin:
Fixes a blocking error

### DIFF
--- a/yaml.properties
+++ b/yaml.properties
@@ -1,7 +1,7 @@
 category=Languages
 description=Allows the analysis of YAML documents
 homepageUrl=https://github.com/sbaudoin/sonar-yaml
-archivedVersions=1.5.2,1.6.0,1.7.0,1.8.0,1.8.1,1.9.0
+archivedVersions=1.5.2,1.6.0,1.7.0,1.8.0
 publicVersions=1.9.1
 
 defaults.mavenGroupId=com.github.sbaudoin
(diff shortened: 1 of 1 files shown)


yaml.properties
```diff
@@ -1,14 +1,20 @@
...
 
+1.9.0.description=Brings 2 new rules and improves some others
+1.9.0.sqVersions=[9.9,LATEST]
+1.9.0.date=2023-11-20
+1.9.0.changelogUrl=https://github.com/sbaudoin/sonar-yaml/releases/tag/v1.9.0
+1.9.0.downloadUrl=https://github.com/sbaudoin/sonar-yaml/releases/download/v1.9.0/sonar-yaml-plugin-1.9.0.jar
+
 1.8.1.description=Fixes an issue that could raise an UnsupportedOperationException when highlighting code
-1.8.1.sqVersions=[9.9,LATEST]
+1.8.1.sqVersions=[9.9,10.2.*]
```
ganncamp, line 17: This range should end with the current version: 10.3

ganncamp (changes requested):

yaml.properties
```diff
@@ -1,15 +1,32 @@
...
+
+1.8.1.description=Fixes an issue that could raise an UnsupportedOperationException when highlighting code
+1.8.1.sqVersions=[9.9,10.3]
+1.8.1.date=2023-11-16
+1.8.1.changelogUrl=https://github.com/sbaudoin/sonar-yaml/releases/tag/v1.8.1
+1.8.1.downloadUrl=https://github.com/sbaudoin/sonar-yaml/releases/download/v1.8.1/sonar-yaml-plugin-1.8.1.jar
 
 1.8.0.description=Brings support for SonarQube 9 LTS and 10; adds a rule and improves some others; fixes some issues
-1.8.0.sqVersions=[9.9,LATEST]
+1.8.0.sqVersions=[9.9,10.2.*]
```
ganncamp, line 29: I think I missed something on my first review. First, this range should (also) end with the current version. Second, there's no need to register and archive 1.8.1 and 1.9.0. There's no point is adding a plugin version to send it straight to the archive and it's acceptable to skip versions/numbers in this file

ganncamp (changes requested):

yaml.properties
sbaudoin: OK, so I skip versions 1.8.1 and 1.9.0.

ganncamp (approved):

PR #481, opened 2023-12-14
sonar-l10-ru 10.2.1

nixel2007:
Hello! Here is a fixed version for 10.2 version of l10n-ru plugin.

10.3 incoming.


nixel2007:
Superseded by #482

PR #498, opened 2024-01-30
archive current version

ganncamp:
Fixes job failure; jar is no longer downloadable.


antoine-vigneau-sonarsource (approved):
All good!! 🙏

PR #530, opened 2024-04-03
Update sonar-dependency-check-plugin to 5.0.0

Reamer:
We've released sonar-dependency-check-plugin version 5.0.0, please add it to the marketplace.
Thanks in advance!

Closes https://github.com/dependency-check/dependency-check-sonar-plugin/issues/911

I want the SonarQube LTS version 9.9.x to continue to receive the old version 4.0.1, is that correct with the two public versions?

SonarCloud-Dashboard: https://sonarcloud.io/dashboard?id=dependency-check_dependency-check-sonar-plugin

### DIFF
--- a/dependencycheck.properties
+++ b/dependencycheck.properties
@@ -2,9 +2,15 @@ category=Integration
 description=Integrates Dependency-Check reports into SonarQube
 homepageUrl=https://github.com/dependency-check/dependency-check-sonar-plugin
 archivedVersions=3.0.1,3.1.0,4.0.0
-publicVersions=4.0.1
+publicVersions=4.0.1,5.0.0
 defaults.mavenGroupId=org.sonarsource.owasp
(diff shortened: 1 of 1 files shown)


Reamer:
CI:
> 10.4 is declared compatible with two public versions of dependencycheck plugin: 5.0.0 and 4.0.1

Yes, SonarQube version 10.4 is compatible with both versions. How can we proceed here?

ganncamp:
@Reamer it's fine to have multiple public versions, but you can't have two public versions that support the same SQ version. So both versions can't be compatible with 10.4. You'll need to end 4.0.1's range at 10.3.

ganncamp (changes requested):
4.0.1's compatibility needs to end at 10.3

Reamer:
> 4.0.1's compatibility needs to end at 10.3

Adjusted

ganncamp (approved):

PR #535, opened 2024-04-12
Fix link to scanner for Gradle release notes to world reachable URL

dorian-burihabwa-sonarsource:


Wohops (approved):
LGTM!
I checked the link with anonymous access and the version is now visible from outside without logging in Jira. Thanks for the fix!

PR #547, opened 2024-05-15
Upgrade JAVA ecoCode plugin to 1.6.1

dedece35:
Hi,

ecoCode JAVA plugin 1.6.1 has been released.

Changes from last SonarQube marketplace integration (1.6.0), please check following changelogs :

from 1.6.0 to 1.6.1 : https://github.com/green-code-initiative/ecoCode/releases/tag/1.6.1
The corresponding SonarCloud project: https://sonarcloud.io/project/overview?id=green-code-initiative_ecoCode

Thank you

Regards


ecocodejava.properties
```diff
@@ -42,3 +42,9 @@ defaults.mavenArtifactId=ecocode-java-plugin
 1.6.0.date=2024-02-02
 1.6.0.downloadUrl=https://github.com/green-code-initiative/ecoCode-java/releases/download/1.6.0/ecocode-java-plugin-1.6.0.jar
 1.6.0.changelogUrl=https://github.com/green-code-initiative/ecoCode-java/releases/tag/1.6.0
+
+1.6.1.description=correction NullPointer in EC2 rule; check Sonarqube 10.5.1 compatibility
+1.6.1.sqVersions=[9.9,10.5.1]
```
ganncamp, line 47: Do you really want to end this at 10.5.1?
If you set it to `LATEST`, it'll work with 10.6 when that's released.
Yes, 10.5 had significant API changes, but that wasn't a normal situation, so there's not usually a need to be super-cautious about forward compatibility. (We do make an effort to warn folks when we're gonna break stuff - on purpose that is :joy: )

ganncamp (changes requested):
See my comment about compatibility. It's ultimately your call, so feel free to resubmit as-is. PR is otherwise okay

ecocodejava.properties
dedece35: Hi @ganncamp
yes, on our ecocode plugins we want to control minimum and maximum compatibility with some tests on each SonarQube version.
For example, there is a problem with python ecocode plugin on 10.5.x, that's why I fix to 10.4.1.
thank for proposition, but we really don't want to be on "latest".
thank you

ganncamp (approved):

PR #581, opened 2024-07-25
Upgrade PHP ecoCode plugin to 1.5.1

dedece35:
Hi,

ecoCode PHP plugin 1.5.1 has been released.

Changes from last SonarQube marketplace integration (1.4.4), please check following changelogs :
https://github.com/green-code-initiative/ecoCode-php/releases/tag/1.5.1

The corresponding SonarCloud project: https://sonarcloud.io/project/overview?id=green-code-initiative_ecoCode-php

Thank you

Regards


ecocodephp.properties
```diff
@@ -35,4 +35,10 @@ defaults.mavenArtifactId=ecocode-php-plugin
 1.4.4.sqVersions=[9.9,10.6.0]
 1.4.4.date=2024-07-19
 1.4.4.downloadUrl=https://github.com/green-code-initiative/ecoCode-php/releases/download/1.4.4/ecocode-php-plugin-1.4.4.jar
-1.4.4.changelogUrl=https://github.com/green-code-initiative/ecoCode-php/releases/tag/1.4.4
\ No newline at end of file
+1.4.4.changelogUrl=https://github.com/green-code-initiative/ecoCode-php/releases/tag/1.4.4
+
+1.5.1.description=incompatible upgrade of dependencies to latest ones; internal technical modifications
```
ganncamp, line 40: What does this mean?

ganncamp (changes requested):
I have a question. Otherwise the details look fine

ecocodephp.properties
dedece35: Hi @ganncamp 
sorry ... maybe not clear ...
these are internal upgrade of all library versions and these upgrade are incompatible modifications to keep a 1.4.X version and it's mandatory to have a 9.9 Sonarqube installed.
maybe, you want me to change the description ?

ganncamp: Well, I was confused because there were some SonarQube API changes early in the 10-series but you declare compatibility with 9.9 through 10.6. :sweat_smile: 
Are there any changes in this version that impact users?

ganncamp (changes requested):

ecocodephp.properties
dedece35: Yes, since 1.5.X, we have upgraded all versions of libraries to be up-to-date. thus, users that uses 1.5.x version of plugin have to upgrade their Sonarqube to minimum version 9.9 and next they can't come back to an older version.
With 1.4.X, users could be roll back to an older one if they wanted even if in description I put 9.9 as minimum version for Sonarqube.
In conclusion, 1.5.X version contains deeper modifications than 1.4.X version.

ganncamp: Is there any user-facing change in this version? I'm not a fan of prompting people to upgrade purely for the purpose of upgrading...

ganncamp (changes requested):

ecocodephp.properties
dedece35: @ganncamp 
for user, no visible modifications, you're right.
but technically speaking, yes, deep modifications and specifically on SonarQube compatibility like explained above.

ganncamp (approved):

PR #600, opened 2024-10-04
Upgrade Python ecoCode plugin to 1.5.0

dedece35:
Hi,

ecoCode Python plugin 1.5.0 has been released.

Changes from last SonarQube marketplace integration (1.4.4), please check following changelogs :
https://github.com/green-code-initiative/ecoCode-python/releases/tag/1.5.0

The corresponding SonarCloud project: https://sonarcloud.io/project/overview?id=green-code-initiative_ecoCode-python

Thank you

Regards


ecocodepython.properties
```diff
@@ -36,3 +36,9 @@ defaults.mavenArtifactId=ecocode-python-plugin
 1.4.4.date=2024-07-18
 1.4.4.downloadUrl=https://github.com/green-code-initiative/ecoCode-python/releases/download/1.4.4/ecocode-python-plugin-1.4.4.jar
 1.4.4.changelogUrl=https://github.com/green-code-initiative/ecoCode-python/releases/tag/1.4.4
+
+1.5.0.description=Deletion of deprecated EC69 rule; support of SonarQube 10.7.0; upgrade internal library versions; technical internal improvements
+1.5.0.sqVersions=[9.9,10.7.0]
```
ganncamp, line 41: Tthe ".0" versions don't actually show up in the Marketplace as ".0". This range should end with either `10.7` or `10.7.*`

That said, I really think you should go with the assumption of compatibility and end these ranges with `LATEST`. I know you're only listing what you've tested, and that could make sense at the start of a cycle (e.g. with 11.0, 11.1, 11.2) but it's generally safe just to use `LATEST` and let your users upgrade SonarQube without losing your functionality.

ganncamp (changes requested):

ecocodepython.properties
dedece35: Hello @ganncamp,
ok thank you for your feedback.
modifications applied.
have a good day !

ganncamp (approved):

dedece35:
Hi @ganncamp, thank you !

PR #605, opened 2024-10-23
AZDO extensions release 3.0.2&7.0.2

denis-troller:


scannerazuresc.properties
```diff
@@ -11,7 +11,7 @@ publicVersions=3.0.2
 
 3.0.2.description=Fix .NET Framework scanner embedding logic.
```
farah22: would it be the same description than above? Fix windows path parsing coming from predefined variables correctly in extraProperties cc @denis-troller

farah22 (approved):
All in order. Thanks!

PR #609, opened 2024-10-29
Update l10nzhtw.properties #603

JE-Chen:
#603

### DIFF
--- a/l10nzhtw.properties
+++ b/l10nzhtw.properties
@@ -1,14 +1,20 @@
 category=Localization
 description=SonarQube Traditional Chinese Language Pack (not compatible with the "Chinese Pack" which uses Simplified Chinese)
-homepageUrl=https://github.com/timlee/sonar-l10n-zh-tw
+homepageUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw
 archivedVersions=
-publicVersions=1.0
+publicVersions=1.0,1.1
 
 defaults.mavenGroupId=org.codehaus.sonar-plugins.l10n
-defaults.mavenArtifactId=sonar-l10n-zhtw-plugin
+defaults.mavenArtifactId=sonar-l10n-zh-tw-plugin
+
+1.1.description=SonarQube Traditional Chinese Pack Support SonarQube 10.* Version
+1.1.sqVersions=[10.0,LATEST]
+1.1.date=2024-10-21
+1.1.changelogUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/tag/1.1
+1.1.downloadUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/download/1.1/sonar-l10n-zh-tw-plugin.jar
 
 1.0.description=SonarQube Traditional Chinese Pack
-1.0.sqVersions=[9.9,LATEST]
+1.0.sqVersions=[9.9]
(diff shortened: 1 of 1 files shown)


l10nzhtw.properties
```diff
@@ -1,14 +1,20 @@
...
 
 defaults.mavenGroupId=org.codehaus.sonar-plugins.l10n
-defaults.mavenArtifactId=sonar-l10n-zhtw-plugin
+defaults.mavenArtifactId=sonar-l10n-zh-tw-plugin
+
+1.1.description=SonarQube Traditional Chinese Pack Support SonarQube 10.* Version
+1.1.sqVersions=[10.0,LATEST]
+1.1.date=2024-10-21
+1.1.changelogUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/tag/1.1
+1.1.downloadUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/download/1.1/sonar-l10n-zh-tw-plugin.jar
```
ganncamp, line 14: This is 404

ganncamp (changes requested):

l10nzhtw.properties
JE-Chen: Ah, sorry. I changed the jar file, so the old URL returns a 404 error.

JE-Chen: Done

l10nzhtw.properties
```diff
@@ -1,14 +1,20 @@
...
 
 defaults.mavenGroupId=org.codehaus.sonar-plugins.l10n
-defaults.mavenArtifactId=sonar-l10n-zhtw-plugin
+defaults.mavenArtifactId=sonar-l10n-zh-tw-plugin
+
+1.1.description=SonarQube Traditional Chinese Pack Support SonarQube 10.* Version
+1.1.sqVersions=[10.0,LATEST]
+1.1.date=2024-10-21
+1.1.changelogUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/tag/1.1
+1.1.downloadUrl=https://github.com/JE-Chen/sonar-l10n-zh-tw/releases/download/1.1/sonar-l10n-zh-tw-plugin-1.1.jar
```
ganncamp, line 14: Okay, the download works now, but my server won't start 
```
Caused by: java.lang.IllegalStateException: The plugin [l10nzhtw] does not support Java 17.0.12
...
Caused by: java.lang.UnsupportedClassVersionError: org/sonar/plugins/l10n/TraditionalChinesePackPlugin has been compiled by a more recent version of the Java Runtime (class file version 67.0), this version of the Java Runtime only recognizes class file versions up to 61.0
```
Current SonarQube versions support / use Java 17

ganncamp (changes requested):

l10nzhtw.properties
JE-Chen: {75DD9D23-2F8C-4B57-AC18-2E19F979D4BA}
I used the wrong JDK, I will rebuild the jar.

JE-Chen: I have already updated the jar file built using JDK 17.

JE-Chen: Same name & same Url

ganncamp:
I couldn't make my browser switch to Chinese, but I assume you've tested locally, so you're in. Congrats

PR #613, opened 2024-11-07
Update scannermaven.properties

leonardo-pilastri-sonarsource:
I am not sure if we should mention SQ compatibility to be >= 10.6; the scanner wouldn't fail with a lower version, but the provisioning would be disabled.


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #636, opened 2024-12-06
sonar-l10n-zh-24.12

xuhuisheng:
Hi,
sonar-l10n-zh-plugin-10.7 has been released.

The main changes is support SonarQube 24.12.

The Download link and release notes: https://github.com/xuhuisheng/sonar-l10n-zh/releases/download/sonar-l10n-zh-plugin-24.12/sonar-l10n-zh-plugin-24.12.jar

The SonarCloud: https://sonarcloud.io/project/overview?id=xuhuisheng_sonar-l10n-zh

Thank you

Regards

### DIFF
--- a/l10nzh.properties
+++ b/l10nzh.properties
@@ -2,15 +2,23 @@ category=Localization
 description=SonarQube Chinese Pack
 homepageUrl=https://github.com/SonarCommunity/sonar-l10n-zh
 archivedVersions=10.4
-publicVersions=9.9,10.0,10.1,10.2,10.3,10.4.1,10.5,10.6,10.7
+publicVersions=9.9,10.0,10.1,10.2,10.3,10.4.1,10.5,10.6,10.7,24.12
 
 defaults.mavenGroupId=org.codehaus.sonar-plugins.l10n
 defaults.mavenArtifactId=sonar-l10n-zh-plugin
 
+24.12.description=Support SonarQube 24.12
+24.12.sqs=[10.8,LATEST]
+24.12.sqcb=[24.12,LATEST]
+24.12.sqVersions=[24.12,LATEST]
+24.12.date=2024-12-06
+24.12.changelogUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/tag/sonar-l10n-zh-plugin-24.12
+24.12.downloadUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/download/sonar-l10n-zh-plugin-24.12/sonar-l10n-zh-plugin-24.12.jar
+
 10.7.description=Support SonarQube 10.7
-10.7.sqs=[10.8,LATEST]
-10.7.sqcb=[24.12,LATEST]
-10.7.sqVersions=[10.7,10.7]
+10.7.sqs=[10.7,10.7.*]
+10.7.sqcb=[10.7,10.7.*]
+10.7.sqVersions=[10.7,10.7.*]
 10.7.date=2024-10-01
 10.7.changelogUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/tag/sonar-l10n-zh-plugin-10.7
 10.7.downloadUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/download/sonar-l10n-zh-plugin-10.7/sonar-l10n-zh-plugin-10.7.jar
--- a/l10nzh.properties
+++ b/l10nzh.properties
@@ -10,15 +10,12 @@ defaults.mavenArtifactId=sonar-l10n-zh-plugin
 24.12.description=Support SonarQube 24.12
 24.12.sqs=[10.8,LATEST]
 24.12.sqcb=[24.12,LATEST]
-24.12.sqVersions=[24.12,LATEST]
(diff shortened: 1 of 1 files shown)


l10nzh.properties
```diff
@@ -2,15 +2,23 @@ category=Localization
...
+24.12.changelogUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/tag/sonar-l10n-zh-plugin-24.12
+24.12.downloadUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/download/sonar-l10n-zh-plugin-24.12/sonar-l10n-zh-plugin-24.12.jar
+
 10.7.description=Support SonarQube 10.7
-10.7.sqs=[10.8,LATEST]
-10.7.sqcb=[24.12,LATEST]
-10.7.sqVersions=[10.7,10.7]
+10.7.sqs=[10.7,10.7.*]
+10.7.sqcb=[10.7,10.7.*]
+10.7.sqVersions=[10.7,10.7.*]
```
ganncamp, line 21: There's no need for a range here, especially since 10.8 has been released, so we know there won't be a 10.7.1. This will work syntactically, but you could make this just `10.7` (with no brackets) if you wanted.

l10nzh.properties
```diff
@@ -2,15 +2,23 @@ category=Localization
...
+24.12.sqVersions=[24.12,LATEST]
+24.12.date=2024-12-06
+24.12.changelogUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/tag/sonar-l10n-zh-plugin-24.12
+24.12.downloadUrl=https://github.com/xuhuisheng/sonar-l10n-zh/releases/download/sonar-l10n-zh-plugin-24.12/sonar-l10n-zh-plugin-24.12.jar
+
 10.7.description=Support SonarQube 10.7
-10.7.sqs=[10.8,LATEST]
-10.7.sqcb=[24.12,LATEST]
-10.7.sqVersions=[10.7,10.7]
+10.7.sqs=[10.7,10.7.*]
```
ganncamp, line 19: If you really want to end compatibility with 10.7, then there's no need to use the new properties for this version, since `sqs` doesn't work for versions below 10.8.

l10nzh.properties
```diff
@@ -2,15 +2,23 @@ category=Localization
...
+10.7.sqcb=[10.7,10.7.*]
```
ganncamp, line 20: Same as above. If you really want this to end with 10.7, then there's no need for this property.

l10nzh.properties
```diff
@@ -2,15 +2,23 @@ category=Localization
...
-publicVersions=9.9,10.0,10.1,10.2,10.3,10.4.1,10.5,10.6,10.7
+publicVersions=9.9,10.0,10.1,10.2,10.3,10.4.1,10.5,10.6,10.7,24.12
 
 defaults.mavenGroupId=org.codehaus.sonar-plugins.l10n
 defaults.mavenArtifactId=sonar-l10n-zh-plugin
 
+24.12.description=Support SonarQube 24.12
+24.12.sqs=[10.8,LATEST]
+24.12.sqcb=[24.12,LATEST]
+24.12.sqVersions=[24.12,LATEST]
```
ganncamp, line 13: This property is now obsolete. No need to include it in the block going forward since this version's compatibility starts after the split into SonarQube Server and SonarQube Community Build.

ganncamp (changes requested):
I'm requesting minor cleanup of the properties, mainly to prevent confusion going forward.

xuhuisheng:
@ganncamp 
Hi Ann,
  I pushed changes following your advice.

  But I still have a confusion. After 24.12, How can i determine the version of  sqs (SonarQube Server)?

  Where can I find the relationship between 24.12 and 10.8? Seems there won't be a sequence number for sonaqube community build, while Sonarqube server still used sequence number, e.g. 10.8.

ganncamp:
@xuhuisheng there's officially not a relationship between the two. sqcb has a different release cadence, monthly, than sqs, bi-monthly. While the two were released together for this initial split, I'm told that won't be the case going forward. In the months when both are released, I don't know which will come out first, but I would expect them to have approximate parity where the features overlap. 
I haven't tested the limits of the changes to the Marketplace functionality, but I believe it should work to have a plugin version that's only compatible with `sqs` or only with `sqcb` if you want to do that. 
Adding @theshadowco as an FYI since he also (diligently) maintains an l10n plugin.

ganncamp:
Aaaand... Now that I've had you remove `sqVersions` because it's not needed, the build is failing without it. IMO this is a bug on our side, but if I can't get it fixed in a reasonable time I may need you to put that back. :(

xuhuisheng:
Please let me know if need I do anything.
Thank you.

matteo-mara-sonarsource:
> Please let me know if need I do anything. Thank you.

@xuhuisheng can you please rebase from master.
The build error should be fixed now but your branch has to be updated with the latest content from master.

xuhuisheng:
> > Please let me know if need I do anything. Thank you.
> 
> @xuhuisheng can you please rebase from master. The build error should be fixed now but your branch has to be updated with the latest content from master.

Done.

PR #650, opened 2025-01-07
Release SonarQube Community Build 25.1.0.102122

vilchik-elena:


christophelevis (approved):

ganncamp:
@xuhuisheng [this](https://community.sonarsource.com/t/updates-to-sonar-s-community-functionality/130732) might help. We've de-coupled the Community Build releases from Server releases. Community Build will be released monthly and Server every 2 months. In the months were both are released, it's not clear to me yet what order that will happen in or even whether they'll happen at (approximately) the same time.  
The `sqs=` lines declare which versions of SonarQube Server the plugin supports, and the `sqcb=` lines declare which versions of SonarQube Community Build. If you remove the `sqs` property, then your plugin won't be marked as supporting SonarQube Server at all.
If you'd like to discuss this further, can you create a thread in [the community](https://community.sonarsource.com)?

PR #652, opened 2025-01-09
Fix l10nru.properties

vilchik-elena:
Avoid having "LATEST" in 2 versions


vilchik-elena:
@lukasz-jarocki-sonarsource do you think we can have some test to avoid such situations in future?

lukasz-jarocki-sonarsource:
Hi yes,
we could write a code that validates whether LATEST is in 2 versions and fail the pull request. I actually thought this code is already somewhere implenented.

ganncamp:
Yes. This used to fail the generation job. 
And sorry I missed this when I approved/merged the PR

vilchik-elena:
@ganncamp could you approve PR?

l10nru.properties
```diff
@@ -10,14 +10,14 @@ defaults.mavenArtifactId=sonar-l10n-ru-plugin
...
-10.8.sqcb=[24.12,LATEST]
+10.8.sqcb=[25.1,LATEST]
 10.8.downloadUrl=https://github.com/1c-syntax/sonar-l10n-ru/releases/download/v10.8/sonar-l10n-ru-plugin-10.8.jar
 10.8.changelogUrl=https://github.com/1c-syntax/sonar-l10n-ru/releases/tag/v10.8
 
 10.7.description=Support SonarQube 10.7
 10.7.date=2024-11-28
 10.7.sqs=10.7
-10.7.sqcb=[24.12,LATEST]
+10.7.sqcb=[24.12]
```
ganncamp, line 20: @vilchik-elena this shouldn't be a range for one version. Just set it to `10.7.sqcb=24.12`

ganncamp (changes requested):

l10nru.properties
```diff
@@ -10,14 +10,14 @@ defaults.mavenArtifactId=sonar-l10n-ru-plugin
 10.8.description=Support SonarQube 10.8 (SonarQube Core 24.12)
 10.8.date=2025-01-07
 10.8.sqs=[10.8,LATEST]
-10.8.sqcb=[24.12,LATEST]
+10.8.sqcb=[25.1,LATEST]
```
ganncamp, line 13: FYI @theshadowco. Compatibility of public versions can't overlap, so this has been bumped up.

l10nru.properties
```diff
@@ -10,14 +10,14 @@ defaults.mavenArtifactId=sonar-l10n-ru-plugin
...
 10.8.downloadUrl=https://github.com/1c-syntax/sonar-l10n-ru/releases/download/v10.8/sonar-l10n-ru-plugin-10.8.jar
 10.8.changelogUrl=https://github.com/1c-syntax/sonar-l10n-ru/releases/tag/v10.8
 
 10.7.description=Support SonarQube 10.7
 10.7.date=2024-11-28
 10.7.sqs=10.7
```
ganncamp, line 19: FWIW, this line should probably be removed since there is no "sqs" 10.7

ganncamp (approved):

PR #662, opened 2025-01-28
Add support SQ 25.1

theshadowco:
Hello there. communitybsl 1.15.2 is out.

SonarCloud project: https://sonarcloud.io/project/overview?id=1c-syntax_sonar-bsl-plugin-community

@ganncamp, Hi! 

Please, help me specify the versions correctly:
We need it so that version 1.15.1 is used BEFORE 25.1, and 1.15.2 on any, from 9.9 to the latest.

Thank you.

### DIFF
--- a/communitybsl.properties
+++ b/communitybsl.properties
@@ -1,15 +1,23 @@
 category=Languages
 description=Code Analyzer for 1C (BSL)
 homepageUrl=https://1c-syntax.github.io/sonar-bsl-plugin-community/en
-archivedVersions=1.11.0,1.12.0,1.13.0,1.14.0,1.15.0
-publicVersions=1.15.1
+archivedVersions=1.11.0,1.12.0,1.13.0,1.14.0,1.15.0,1.15.1
+publicVersions=1.15.2
 
 defaults.mavenGroupId=com.github.1c-syntax
 defaults.mavenArtifactId=sonar-communitybsl-plugin
 
+1.15.2.description=Support SQ 25.1+
+1.15.2.sqs=[10.8,LATEST]
+1.15.2.sqcb=[24.12,LATEST]
+1.15.2.sqVersions=[9.9,10.7]
+1.15.2.date=2025-01-28
+1.15.2.changelogUrl=https://github.com/1c-syntax/sonar-bsl-plugin-community/releases/tag/v1.15.2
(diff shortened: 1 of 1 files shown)


ganncamp:
I don't understand. Chronologically, 9.9 is _before_ 25.1.

nixel2007:
The question is about sq ranges in the yaml file. Let me provide some context. 

Plugin contains our spring boot application inside. Before SQ 25.1 we invoked "run" on special class marked as spring boot application, and everything worked fined.
Spring version was upgraded in SonarQube between 24.12 and 25.1 and our approach is not working anymore.

New version of plugin (1.15.2) contains a workaround for this issue, we do not depend on spring boot application anymore. So plugin version 1.15.2 works well on SQ any version in range 9.9-LATEST (INCLUDING 25.1).
What we want to achieve is limiting version 1.15.1 to [9.9, 24.12], so users who run SQ prior to 25.1 will be able to upgrade communitybsl plugin to 1.15.2 via Marketplace first, and then do an upgrade to SQ 25.1 (or newer) .

So Valeriy asks if he made changes in the yaml file in a correct way. Hope it helps.

ganncamp:
Are you saying you don't want any new users to install 1.15.1? If so, then you accomplish that by archiving it.

ganncamp (approved):
I'm approving this, but I won't merge it until I get confirmation that this is what you want: 
* hide 1.15.1 from new users
* publish 1.15.2 to all new users
* make 1.15.2 available to existing users for upgrade

nixel2007:
We want to hide 1.15.1 only on SQ 25.1+. 

>hide 1.15.1 from new users
publish 1.15.2 to all new users
make 1.15.2 available to existing users for upgrade

^ seems correct. Is current setup in yaml correct for this? 
Since one can't install previous version of the plugin from marketplace, archiving of 1.15.1 should solve the main issue. however AFAIR this yaml file is used for compatibility matrix as well, and we would like to highlight that versions prior to 1.15.2 are not compatible with SQ 25.1

ganncamp:
>would like to highlight that versions prior to 1.15.2 are not compatible with SQ 25.1

There's no real "highlight" available. Just publishing. I'm going to merge this and run the job. Come back to me (with a new PR?) if what comes out doesn't look right to you

nixel2007:
Thanks, Ann!

PR #663, opened 2025-01-29
REL-3507 Add 2025.1 downloadUrl for <=10.7 compatibility

matteo-mara-sonarsource:
Validate locally by reproducing the error without the downloadUrl available when running a 10.7 Community Edition.

By adding the property in the file the modal gets displayed properly without JS errors:

image

https://github.com/user-attachments/assets/7466e0ae-9677-44a6-b550-0d6ec4f690c9

### DIFF
--- a/update-center-source.properties
+++ b/update-center-source.properties
@@ -13,6 +13,9 @@ ltaVersion=2025.1
 pastLtaVersion=9.9.8
 
 2025.1.description=Long-Term Active version - advanced AI capabilities, many security breakthroughs, enhanced developer experience, simplified operability, many new language features, and much more
+# "downloadUrl" needed to notify users of SQ<=10.7 of new releases linking to the general download page since there is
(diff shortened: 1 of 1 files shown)


christophelevis:
Thanks @matteo-mara-sonarsource. It seems to temporarily fix the problem.
What about when 25.2 will be released? Does the version of CB matter here, and is it displayed somewhere?

matteo-mara-sonarsource:
> Thanks @matteo-mara-sonarsource. It seems to temporarily fix the problem. What about when 25.2 will be released? Does the version of CB matter here, and is it displayed somewhere?

@christophelevis no we don't display it. We need it for retro-compatibility. I pointed to 25.1 just for consistency with what was done when 10.8 was released. 
The acceptance criteria document suggests to just point to binaries.sonarsource.com.

christophelevis (approved):
Looks good. Thanks!

PR #688, opened 2025-03-14
Release FindBugs 4.4.2

gtoison:
We've released [sonar-findbugs-plugin 4.4.2](https://github.com/spotbugs/sonar-findbugs/releases/tag/4.4.2), please add it to the marketplace.
Detailed changelog: https://github.com/spotbugs/sonar-findbugs/releases/tag/4.4.2
Download URL: https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.4.2/sonar-findbugs-plugin-4.4.2.jar
SonarCloud: https://sonarcloud.io/summary/new_code?id=com.github.spotbugs:sonar-findbugs-plugin

Thanks in advance!


findbugs.properties
```diff
@@ -57,9 +57,16 @@ defaults.mavenArtifactId = sonar-findbugs-plugin
...
-4.4.0.sqs=[10.8,LATEST]
-4.4.0.sqcb=[24.12,LATEST]
+4.4.0.sqs=[10.8,2025.1.*]
+4.4.0.sqcb=[24.12,25.3.*]
 4.4.0.sqVersions = [9.9,10.7]
 4.4.0.date = 2025-02-14
 4.4.0.changelogUrl = https://github.com/spotbugs/sonar-findbugs/releases/tag/4.4.0
 4.4.0.downloadUrl = https://repo.maven.apache.org/maven2/com/github/spotbugs/sonar-findbugs-plugin/4.4.0/sonar-findbugs-plugin-4.4.0.jar
+4.4.2.description = Use SpotBugs 4.9.1, sb-contrib 7.6.9, and findsecbugs 1.13.0
+4.4.2.sqs=[10.8,LATEST]
```
ganncamp, line 67: Now that you've caught up on the parameters, I have to question you on the compatibility. Your changelog says you've reverted the SQ dependency to be compatible with the _previous_ LTA, 9.9.8. If that's the case, then your 4.4.2 version is likely not compatible past SonarQube 10.2 (altho I'm too lazy to test). 
I suggest that you reconsider this reversion.
In the end it's up to you, but if you really want to revert, then we cannot publish this plugin version as being compatible with 10.2 and higher.

ganncamp (changes requested):

hazendaz:
@gtoison Are you sending these manually or is this happening automatically on your account too?

findbugs.properties
gtoison: The compatibility issue occured because the `sonar-plugin-api` dependency was upgraded to `9.17.0.587` by mistake, this cause the sonar packaging plugin to add this as the minimum supported in the manifest of the plugin.
Reverting to `9.14.0.375` makes the plugin compatible again with SonarQube 9.9.x because there was no reason to require `9.17.0.587`.
The integration tests of the plugin currently test it against SonarQube 9.9.7, 10.7 and 25.2.0 so I think the compatiblity is ok (I'm not sure what you mean when you say that it might not be compatible with 10.2).

gtoison: I've also updated the integration tests to run against SonarQube 9.9.9 and 25.3.0 so the plugin should be compatible with these versions as well

ganncamp (approved):

PR #703, opened 2025-04-01
Comment out unavailable sonar-l10n-zh-tw-plugin-1.1

vilchik-elena:
cc @JE-Chen


ganncamp (commented):
The build fails immediately

ganncamp (approved):
Looks like the build failure is in update-center-source.properties so I guess this PR is okay

PR #714, opened 2025-05-06
Update to SCMTFVC 2.5.0

lg2de:


scmtfvc.properties
```diff
@@ -2,12 +2,19 @@ category=Integration
...
+2.5.0.description=update DevOps api
+2.5.0.sqs=[10.8,LATEST]
+2.5.0.sqcb=[24.12,LATEST]
+2.5.0.downloadUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/download/2.5.0/sonar-scm-tfvc-plugin-2.5.0.318.jar
+2.5.0.date=2025-05-05
+2.5.0.changelogUrl=https://github.com/lg2de/sonar-scm-tfvc/releases/tag/2.5.0
+
 2.4.0.description=update DevOps api
 2.4.0.sqVersions=[9.9,10.7]
 2.4.0.sqs=[10.8,LATEST]
```
ganncamp, line 20: This and the next line need to be changed. Archived versions can't be compatible with `LATEST`, and only one version can be compatible with `LATEST` at a time. You should set these end ranges to the latest versions, currently 2025.2 for `sqs` and 25.5 for `sqcb`

ganncamp (changes requested):
In addition to the specific changes I've requested, I need you to point me to the SonarQube Cloud analysis of this project.

scmtfvc.properties
lg2de: Even I do not fully understand, why older versions cannot be compatible with LATEST (because they are!), I've changed this.

lg2de:
The sonar analysis can be found here: https://sonarcloud.io/project/overview?id=lg2de_sonar-scm-tfvc

ganncamp:
FYI, the mechanics of the metadata generation job are what only allow one, `public` version to be compatible with `LATEST` at a time. The job will fail if the condition isn't met.

lg2de:
> FYI, the mechanics of the metadata generation job are what only allow one, `public` version to be compatible with `LATEST` at a time. The job will fail if the condition isn't met.

You explained, that the mechanism is build like that.
But, you did explain, why this restriction is useful.
Older versions ARE compatible with current SQ. So, why specifying it is not compatible?

ganncamp:
Okay, first, yes this is slightly arbitrary.

The rules hard-coded into the job include:
* `archived` versions can't be compatible with `LATEST`
* only one version, total, can be compatible with `LATEST`

The underlying goal is to help SQ admins manage their plugins and prompt them to install or upgrade to the most appropriate version.

> Older versions ARE compatible with current SQ. So, why specifying it is not compatible?

Because we don't want them to _install_ older versions. We all (you as a plugin maintainer, and me as a SonarSourcer) want them to install the latest/greatest

PR #731, opened 2025-06-05
Sonar jQAssistant Plugin 3.0.1

DirkMahler:
We noticed that our latest 3.0.0 release was sadly broken, there we created this 3.0.1 bug fix.

We would like to remove 3.0.0 completely from the market place, hopefully this PR does it correctly.

Latest SQ report is here: https://sonarcloud.io/project/overview?id=jqassistant-tooling_sonar-jqassistant-plugin


ganncamp:
You really don't want to remove the old version entirely; then people won't be prompted to upgrade. Just move the old version to archived & publish the new one

ganncamp (changes requested):
Can't request changes without leaving a comment

> You really don't want to remove the old version entirely; then people won't be prompted to upgrade. Just move the old version to archived & publish the new one

DirkMahler:
I re-introduced the 3.0.0 release, hopefully I got the right version ranges.

PR #739, opened 2025-06-17
Add magento2php plugin

rostilos:
Hi.
PR is proposing to add a plugin with a set of rules for Magento 2 platform.
Thread on the community forum :


magento2php.properties
```diff
@@ -0,0 +1,15 @@
...
+homepageUrl=https://github.com/rostilos/sonarqube-magento2-rules
+archivedVersions=
+publicVersions=1.3.0
+
+defaults.mavenGroupId=org.rostilos.sonarrules
+defaults.mavenArtifactId=sonar-magento2-rules
+
+1.3.0.description=PHP Custom Rules for Magento 2 CMS
+1.3.0.sqs=[10.6,24.12]
+1.3.0.sqcb=[24.12,24.12]
```
ganncamp, line 12: I've just tested this on 25.5 (haven't gotten around to downloading 25.6 yet). You should set this end range to `LATEST`, like so
`1.3.0.sqcb=[24.12,LATEST]`

magento2php.properties
```diff
@@ -0,0 +1,15 @@
...
+description=Adds rules for Magento2 PHP development
+homepageUrl=https://github.com/rostilos/sonarqube-magento2-rules
+archivedVersions=
+publicVersions=1.3.0
+
+defaults.mavenGroupId=org.rostilos.sonarrules
+defaults.mavenArtifactId=sonar-magento2-rules
+
+1.3.0.description=PHP Custom Rules for Magento 2 CMS
+1.3.0.sqs=[10.6,24.12]
```
ganncamp, line 11: This `sqs` argument declares the plugin version's compatibility with SonarQube server, the versioning of which now looks like 2025.1 ... So this should be something like 
`1.3.0.sqs=[2025.1,LATEST]`
Note that I suggest starting this at the current LTA rather than the EOL 10.6 version since this is a brand new plugin, so you can't already have users on that older version

ganncamp (changes requested):

ganncamp (approved):

Metaorweorusorour:
> Hi. PR is proposing to add a plugin with a set of

PR #749, opened 2025-07-11
checkstyle: update to 10.26.1

muhlba91:


ganncamp:
I don't really understand the comment you've linked to.

And a code change in the plugin was required to do the upgrade in this version. That should be analyzed on SonarQube Cloud.

PR #768, opened 2025-08-19
OPE-843 Moving to github actions

lukasz-jarocki-sonarsource:


alain-kermis-sonarsource (approved):
Nice work, let's keep an eye out for when the time comes to remove all references to Cirrus.

PR #773, opened 2025-08-26
Fixing typo in CODEOWNERS

julienlancelot:
Fixing code owners typo

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-.github/CODEOWNERS @Sonarsource/platform-onprem-squad
+.github/CODEOWNERS @SonarSource/platform-onprem-squad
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-.github/CODEOWNERS @sonarsource/platform-onprem-squad
+.github/CODEOWNERS @Sonarsource/platform-onprem-squad
```
matteo-mara-sonarsource, line 1: If it is case sensitive I think it is supposed to be `SonarSource`.

julienlancelot: Arg yes indeed

matteo-mara-sonarsource (approved):
LGTM

